PR #49, opened 2015-09-23
[WIP] Add mul_ function for polynomial multiplications

tueda:
Work in progress. Maybe I need someone's review, because I'm not 100% sure about what I'm doing and maybe I missed something. Please don't merge this PR for now. (But it's nice to see the difference of sources on GitHub easily, isn't it?)

I'm doing some big polynomial manipulations. Now the bottleneck seems to be in repeated operations like

```
  #$x = $a * $d - $b * $c;
```

and presumably in the multiplications in it. 

(Actually I got some errors during this big computation

```
Attempt to write 32768 bytes on file 4 at position      1006337560
RetCode = -1, Buffer address = 47495421420296
Called from MergePatches with k = 7 (stream 3)
Called from StoreTerm
Program terminating at externalchannel Line 1 -->
```

Here `externalchannel Line 1` comes from the fact that I'm using external channels. But this crash may be another story. I'm rerunning the same program to see if it came from an OS error (disk full, disk io, pipe io, etc.))

Because the $-variables are just multivariate polynomials, I thought maybe this can be speeded up by the polynomial multiplication using heap in `poly.cc`, though it would take some time for conversions. (If you have another idea, it's welcome.) So I made this experimental patch to add `mul_(a,b)` function. A benchmark test is at https://gist.githubusercontent.com/tueda/4f1ddd75e7cd968f975c/raw/664ea4b8578ea8bf398052b80ccddde79778f413/test-polymul.frm. It compares `$a*$b` and `mul_($a,$b)`. The result is as follows (N is the size of the problem):

```
N = 5
~~~method A, w/o mul_ : .13 sec
~~~method B, w/ mul_ : .14 sec

N = 6
~~~method A, w/o mul_ : .82 sec
~~~method B, w/ mul_ : .71 sec

N = 7
~~~method A, w/o mul_ : 4.36 sec
~~~method B, w/ mul_ : 3.30 sec

N = 8
~~~method A, w/o mul_ : 18.71 sec
~~~method B, w/ mul_ : 13.10 sec

N = 9
StageSort
StageSort
StageSort
~~~method A, w/o mul_ : 78.89 sec
~~~method B, w/ mul_ : 47.06 sec
```

So the speedup factor becomes visible when the polynomial size is large.

For now, I'm setting `sort_vars=false` and `sort_univar=true` in the conversion to polynomials, but this choice might be for this specific benchmark program.


tueda:
Another benchmark test that I found: The Fateman benchmark.
```
Symbols x,y,z;
Local p = (1+x+y+z)^20;
.sort
Local q = p+1;
.sort
Local r = p*q;
.end
```
takes 1.27 seconds on my machine, while
```
Symbols x,y,z;
#$p = (1+x+y+z)^20;
#$q = $p+1;
#$r = mul_($p,$q);
L r = $r;
.end
```
takes only 0.12 seconds. For a comparison, the following Mathematica code
```
Timing[
  p = Expand[(1+x+y+z)^20];
  q = p + 1;
  r = Expand[p*q];
]
```
takes 0.21 seconds on the same machine (v11.0.1).

PR #91, opened 2016-05-04
Additional sort info and scratch usage

jodavies:
I found these tweaks useful when trying to correctly configure the buffer parameters for large memory machines, in order to not touch the disk.

(I know MaxScratchSize is more or less what minos extracts from the log file...)

Possibly helpful to include bits?

Josh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ file "[AUTHORS](AUTHORS)".
 Build instructions
 ------------
 
-Before building FORM, it is adviced to install the optional dependencies `gmp` and `zlib` for better performance. To quickly build FORM, install the `autoconf` pacakge. Then run:
+Before building FORM, it is advised to install the optional dependencies `gmp` and `zlib` for better performance. To quickly build FORM, install the `autoconf` package. Then run:
 
     autoreconf -i
     ./configure


benruijl:
The patch is looking good. Any comments from the others? Perhaps there is even more size information we would like to print?

tueda:
1. This is just a question for consistency of the code, but why do we need to use `printf` instead of `MesPrint`? I guess `MesPrint` always flushes the console output. For `MesPrint`, locking `ErrorMessageLock` may be needed for TFORM and ParFORM. I'm not sure what could happen for `printf` in the parallel versions.
2. I guess "Max. space for InParallel" is always zero for FORM and ParFORM, and could be switched off for them. (Or better to have it always?)
3. All "StageSort" messages would also have "SORTINFO: " header. It makes grepping this kind of information from log files easier.

jodavies:
Printing the number of patches being merged in stage4 is not as helpful as anticipated. It looks like threads perform a stage4 sort *as soon as* the number of file patches hits FilePatches, whether it has finished processing a module's input or not.

You can't, therefore, look at the number printed and set your FilePatches accordingly.

PR #101, opened 2016-06-10
[WIP] Add substr_ macro

tueda:
Although this may be a "feature creep" and I'm not sure whether many people find it useful or not, I would like to make a "Work in Progress" pull request, hoping this may stimulate other discussions.

As far as I know, there is no way/trick to extract a part of a string in FORM. So I experimentally implemented `substr_(STRING,POS,LEN)` macro, which is more or less `substr` of Perl and PHP but the first character is at POS=1. (Note that it is not `substring_(STRING,FROM,TO)`). `POS` and `LEN` are somewhat extended to non-positive and negative integers, respectively. Especially, `LEN=-1` means the actual length of `STRING`.

An application is defining a procedure with optional parameters:
```
#procedure Proc1(x,?opts)
  #define expand "-1"
  #do opt={`?opts'}
    #ifdef `opt'
      #switch `opt'
        #case expand
          #redefine expand "0"
          #break
        #case noexpand
          #redefine expand "-1"
          #break
        #default
          #if "`substr_(`opt',1,7)'" == "expand="
            #redefine expand "`substr_(`opt',8,-1)'"
          #else
            #message Error: illegal option of Proc1: `opt'
            #terminate
          #endif
          #break
      #endswitch
    #endif
  #enddo

  #message x=`x'
  #message expand=`expand'
#endprocedure
```
which accepts `#call Proc1(x,expand=2)` as well as `#call Proc1(x)`, `#call Proc1(x,expand)` or `#call Proc1(x,noexpand)`.

This branch also contains an implementation of `strlen_(STRING)`, but it is not primitive in the sense that it can be implemented based on `substr_`:
```
#procedure StrLen(?str,dest)
  #redefine `dest' "0"
  #do loop=1,1
    #if "`substr_(`?str',1,``dest'')'" != "`?str'"
      #redefine `dest' "{``dest''+1}"
      #redefine loop "0"
    #endif
  #enddo
#endprocedure

#define n ""
#call StrLen(12345,n)
```

Other string operations like `strreplace_(STRING,OLD,NEW)` can also be implemented based on `substr_`.

### DIFF
--- a/sources/pre.c
+++ b/sources/pre.c
@@ -338,6 +338,82 @@ higherlevel:
 						PutPreVar(namebuf,(UBYTE *)buf,(UBYTE *)"?a",1);
 						goto dostream;
 					}
+					else if ( StrICmp(namebuf,(UBYTE *)"substr_") == 0 ) {
+						/*
+						 * `substr_(STRING,POS,LEN)' gives the substring of
+						 * STRING starting from POS (the first character is
+						 * at POS=1) and with the length of LEN.
+						 */
+						LONG len = i-(s-namebuf)-2;
+						/*
+						 * Find two "," from the end.
+						 */
+						UBYTE *ss1, *ss2;
+						ss2 = s+len-1;
+						while ( ss2 >= s && *ss2 != ',' ) ss2--;
+						ss1 = ss2-1;
+						while ( ss1 >= s && *ss1 != ',' ) ss1--;
+						if ( ss1 < s ) {
+							/*
+							 * Not found.
+							 */
+							goto substr_error;
+						}
+						else {
+							/*
+							 * Parse the last two parameters as numbers.
+							 */
+							LONG n1, n2;
+							len = ss1-s;
+							*ss1++ = 0;
+							*ss2++ = 0;
+							if ( *ss1 == 0 || *ss2 == 0 ) {
+								goto substr_error;
+							}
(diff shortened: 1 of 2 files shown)


tueda:
FYI: if your pull request is still incomplete and needs more work, then there is the "Draft" feature for pull requests in GitHub.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

Even if your work is in progress, creating a draft pull request to get others' feedback sometimes be beneficial (I have changed the status of this pull request to "Draft").

tueda:
Note: I found https://github.com/vermaseren/form/commit/1d4b775308c457c9a8ca72aa64890ac9642ec4dd added `takeleft_`, `takeright_`, `keepleft_` and `keepright_`.

Example:
```form
#message `takeleft_(abcdefg,2)'
#message `takeright_(abcdefg,2)'
#message `keepleft_(abcdefg,2)'
#message `keepright_(abcdefg,2)'
.end
```
```
    #message `takeleft_(abcdefg,2)'
~~~cdefg
    #message `takeright_(abcdefg,2)'
~~~abcde
    #message `keepleft_(abcdefg,2)'
~~~ab
    #message `keepright_(abcdefg,2)'
~~~fg
```

PR #123, opened 2016-09-01
Added support for folds in the same file

apik:
Can be used as:
    #include # foldname

Here is a code example (to be placed in file `unnamedfold.frm`):

```
#-

#include unnamedfold.frm # f1

#include # f1

.end

*--#[ f1 :
        #message fold included
*--#] f1 :
```


tueda:
It looks like a syntax sugar of
```
#include `NAME_' # f1
```

apik:
I agree with you and going to close  pull request, just forgotten about preprocessor variables.

Issue #264, opened 2018-02-09
stdin as the input file?


tueda:
Suppose one wants to run some FORM program that is dynamically created in a shell script or some program. Then one has to make a temporary file for the FORM input. The file name should be unique and the file must be guaranteed to be deleted after finishing the calculation while FORM needs to access to it by the file name (i.e., `unlink` before `close` doesn't work). This is a bit cumbersome and one is forced to handle
1. making a unique file name for a temporary file (or fifo) in a secure way,
2. signal handling for deleting the file; deleted even when `Ctrl+C` is pressed.

`mkstemp` to generate a unique name `tmpXXXXXX` or  process-private pipes for  `/dev/fd/<n>` are useless because FORM doesn't accept file names without `.frm`. Signal handlers are basically global, so a pain in the neck for OOP.

So I was just wondering why FORM doesn't support input passed via the standard input, which makes such temporary files unnecessary. It would be nice if the following could work:
```shell
cat <<END | form -q -i
Off stats;
S x;
L F = (1+x)^2;
P;
.end
END
```
Here I have assumed `-i` indicates FORM reads input from the standard input (though I know it is reserved for "interactive" session). The option may be other forms, e.g., `-in`, or we could make a special file name `-` (as `form -`) that means the standard input.

tueda:
What would be wrong if FORM accepts input file names without `.frm`?

As I wrote above, FORM rejects when I pass the file name `/dev/fd/5` and says `/dev/fd/5.frm` doesn't exist, even if `/dev/fd/5` certainly exists and `/dev/fd/5.frm` never exists. Another example is a script as an executable. The following script
```
#-!bin/sh
#- \
exec form -q "$0"
L F = 1;
P;
.end
```
can be executable if it is named as `foo.frm` with `chmod +x foo.frm`. But it doesn't work if I rename it to `foo`.

vermaseren:
A couple of remarks.
Form scans the input file for instructions that may refer to a setup.
Then it rewinds the input and after allocations have been made, it really starts.
This makes stdin rather tricky.
Version one and two did not need the .frm extension. It made things at times rather
chaotic. Hence at a certain moment it was decided that the extension .frm would
be mandatory. It is possible to make an option in the command tail that tells that
the file does not have the extension, just for very special use.
Note that the C compiler also insists on the extension .c etc.

I think it is good that there are conventions. And whatever convention one
chooses, there will always be cases in which they may not be optimal.

Jos

tueda:
[...]

PR #265, opened 2018-02-10
Read the input from stdin instead of a .frm file

tueda:
Work in progress. This is an experimental implementation of #264. Example:
```shell
cat <<END | form -q -stdin
Off stats;
S x;
L F = (1+x)^2;
P;
.end
END
```
Setup parameters (`#:...`) are not supported yet.

It also works with Windows (an binary available in AppVeyor: https://ci.appveyor.com/project/tueda/form/build/job/5ndaxgejblwqa4ho/artifacts).

### DIFF
--- a/sources/startup.c
+++ b/sources/startup.c
@@ -407,6 +407,7 @@ printversion:;
 							errorflag++;
 #endif
 							AM.FromStdin = 1;
+							AC.NoShowInput = 1; // disable input echoing by default
 							break;
 				default:
 						if ( FG.cTable[*s] == 1 ) {
@@ -1603,6 +1604,7 @@ int main(int argc, char **argv)
 	 */
 	AS.printflag = -1;
 #endif
+	PrintHeader(1);
 
 	if ( ( retval = DoTail(argc,(UBYTE **)argv) ) != 0 ) {
 		if ( retval > 0 ) Terminate(0);
@@ -1634,7 +1636,6 @@ int main(int argc, char **argv)
 	ReserveTempFiles(0);
 	IniFbuffer(AT.fbufnum);
 #endif
-	PrintHeader(1);
 	IniVars();
 	Globalize(1);
 	if ( AM.TimeLimit > 0 ) alarm(AM.TimeLimit);


tueda:
Now setup parameters are supported. Example:
```shell
cat <<END | form -stdin
#:workspace 1234K
#:wtimestats on
On setup;
S x;
L F = (1+x)^2;
P;
.end
END
```

tueda:
Squashed into one commit. Ready to merge (once the others agree).

Possible improvement: currently the stdin is read one by one, for simplicity, by `ReadFile(stream->handle,&c,1)` in `ReadFromStream`.

alexmyczko:
Normally for stdin as file input - is used, if - must be given as parameter to program it's -- -.

Can -stdin be replaced by - for consistency among other command line software? so it should just be
form -

tueda:
By the way, the `PADXXX` macros are useless, memory inefficient, and moreover **hard to merge** two changes! (Now this branch is non-trivially conflicting with the master...)

tueda:
Now the filename `-` means stdin.

benruijl:
Is this change still ready to be merged?

tueda:
It seems that there are no conflicts with the current master branch (though the current master fails to pass the tests and travis-ci.org is dead), but probably it has conflicts with 5.0 via `PADXXX`, I guess.

benruijl:
I've changed the pull request so that the banner is printed first and the line echoing is disabled. Now form can be used pretty cleanly from the command line as if it were an interpreter! I imagine this may be useful when teaching form.

@tueda can you check if moving the header print is safe for parFORM? If so ,we can merge it :)

tueda:
We can ignore ParFORM for now because MPI doesn't automatically broadcast stdin (see, for example, https://stackoverflow.com/q/49016638) so it doesn't work anyway. I have added an error message for the case where `-` is specified in ParFORM.

You can use "Rebase and merge" (or "Squash and merge" if you want).

benruijl:
I've moved the printing of the header, even when we don't use stdin mode. I think it's safe.

jodavies:
[36f5536](https://github.com/vermaseren/form/pull/265/commits/36f5536083fc8f4c4158daada677702c05499a03) causes tform's header to always state "-1 workers", by the way.

tueda:
`git revert 36f5536` seems to fix "-1 workers".

tueda:
If `PrintHeader` is called before `DoTail`, then any command-line options like `-w4` are not processed yet, so `tform` has to print `-1 workers`, I guess.

jodavies:
Actually, this also breaks the `-q` option, the header prints regardless.

tueda:
Maybe a solution would be:
1. Revert the position of `PrintHeader(1)` back to just before `IniVars()`, and
2. Use `AM.silent = 1` (as the `-q` option) instead of `AC.NoShowInput = 1` for `-` (stdin),

though I'm not sure what is the exact differences between `AM.silent = 1` and `AC.NoShowInput = 1`.

tueda:
A safer option:
1. Revert `PrintHeader(1)`,
2. Keep `AC.NoShowInput = 1` for `-` (stdin), and
3. Skip `PrintHeader(1)` if `AM.FromStdin` is `1`.

By the way, if FORM always doesn't print the version header for the stdin case, how can one know the exact version of FORM running? Currently, parsing the version header is the only way to know that.

alexmyczko:
form has a -v option to only print its version?

benruijl:
@tueda I'll push your latest suggestion. I've tested it and it seems to work. Running form without any arguments or with `-v` still prints the version.

tueda:
> form has a -v option to only print its version and exit?

Yes, indeed this seems to be enough. The user program may run FORM twice, `form -v` to check the version and then `form -` for the main program. (For [`python-form`](https://github.com/tueda/python-form), I wrote code checking the version and running the main program in [one-go](https://github.com/tueda/python-form/blob/24a77ca8e904a7e45c8c07b627dfa39b553e2011/form/formlink.py#L157).)

> I'll push your latest suggestion. I've tested it and it seems to work. Running form without any arguments or with -v still prints the version.

Nice! Go ahead.

PR #276, opened 2018-03-30
Replace ftime() by clock_gettime()

mreininghaus:
`ftime()` is considered obsolete in the POSIX standard and not available anymore on some platforms, e.g. OpenBSD. Therefore, I replaced it by `clock_gettime()`, however keeping `ftime()` as fallback.


tueda:
Thank you for your pull request! I totally agree with you that `ftime()` is obsolete and it should be replaced at some point. Some comments:
- Probably checking `clock_gettime()` in the `configure` is needed. Some system may require linking to `rt` and this is the reason why the CI failed. Once the check is done via `configure`, `_POSIX_TIMERS` is replaced with `HAVE_CLOCK_GETTIME` or something like that. For reference, my attempt for `gettimeofday()` is [here](https://github.com/tueda/form/commit/858b2c58bf3d966dbcb8374977bcacd2f95e1b33).
- Because this is a set of rather small patches, would you mind squashing them into one commit and rebasing to the master such that fast-forward merge would work?

mreininghaus:
Thank you for your comments. I'll implement your suggestions.

tueda:
@mreininghaus Sorry that I didn't look into this for quite a while.

Do you still want to edit your commit history? Especially your 3 merge commits nonlinearly complicates the commit graph.

If you don't mind, I could git-cherry-pick your commits without the merge commits. The result will be like https://github.com/tueda/form/commits/mreininghaus-ftime-clock_gettime and this can be merged into the master branch.

mreininghaus:
Feel free to do that. Sorry if I messed things up a bit.

tueda:
OK, I will try to merge it in that way. Thanks for your contribution!

Issue #112, opened 2016-07-26
Filename collisions in sort files


jodavies:
Now that the sort files can be put in a different directory to the scratch files, (TempDir != TempSortDir), it is not safe to assume that sort files can be created with the same base filename.

For eg, if I start tform on a script which creates sort files with
#: TempDir /foo1

and another with
#: TempDir /foo2
#: TempSortDir /foo1

the second instance of tform will overwrite the existing xformxxx.sFN of the first, causing it to crash when it comes to merge the patches.

I suppose one way to avoid this is to always use -M, but this is not the default behaviour. Otherwise, threads must check each and every time they create a sort file.

Thanks,
Josh.

vsht:
I'm kind of wondering how unique the sort files become when the -M option is used? Is there still a chance that when
running multiple tform processes and setting `FORMTMP` to `/tmp` a crash due to overwriting may occur?

I came up with my own workaround by using sth like
```
tfmTformTmpDir="$(mktemp -d)";
cleanup="yes"
tform -t "$tfmTformTmpDir" ...;
rm -rf $tfmTformTmpDir
```
which creates unique subfolders for each tform process. However, if there is a reliable built-in mechanism available,
I'd rather use that than reinventing the wheel.

jodavies:
`-M` means the filename will use the first 5 digits of the `pid`. It is unlikely that there will be a collision, but I think it is not guaranteed.

vsht:
I see, thanks. Then I'll try using that and see if it suits me. On the cluster this is never an issue, since each tform
process gets a unique environment, but on my laptop (Ryzen 4750U with 8C/16T) I have to take some precautions by hand.

vermaseren:
In the -M option the temporary files have a name that contains the PID of the running (t)form.
This PID is supposed to be unique, UNLESS you have old files that were not removed
with an old PID in them of a job that ran some time ago. But that would mean that you loose
those old files. Normally that is no disaster, but in principle old .str files can be used in a
Load statement to recover their contents.

In the scheme without the PID, Form checks whether a .str file with the selected name exists
already. If so it raises the number in the name and tries again. This is also not failsafe, because
when you start programs very quickly after each other the .str file of the first job may not exist
yet when the second looks for that name. Also, sometimes when you delete the storage, the
file is temporarily removed and recreated.

jodavies:
Indeed, on clusters usually there are no issues. On your laptop you would have to have two tform processes running whose `pid` differs by precisely 100000.

[...]

PR #284, opened 2018-06-19
Append pid to spectator filename, if form -M

jodavies:
Make spectators nicer to use when running multiple concurrent jobs. We don't have to guarantee unique temp directories with the -M option.

### DIFF
--- a/sources/spectator.c
+++ b/sources/spectator.c
@@ -123,12 +123,24 @@ int CoCreateSpectator(UBYTE *inp)
 	while ( *q && ( *q == ',' || *q == ' ' || *q == '\t' ) ) q++;
 	if ( *q ) goto Syntax;
 	cc = *p; *p = 0;
+
+/*
+	Append pid if AM.MultiRun. Allocate space for supplied name + '.' + 5 digits + '\0'
+*/
+	char *filename2 = Malloc1(sizeof(filename) + 1 + 1 + 5, "MultiRun Spectator filename");
+	if ( AM.MultiRun ) {
+		sprintf(filename2, "%s.%d", filename, ((int)GetPID())%100000);
+	}
+	else {
+		sprintf(filename2, "%s", filename);
+	}
 /*
 	Now we need to: create a struct for the spectator file.
 */
 	if ( HadOne == 0 )
 		numexpr = EntVar(CEXPRESSION,inp,SPECTATOREXPRESSION,0,0,0);
-	fh = AllocFileHandle(1,(char *)filename);
+	fh = AllocFileHandle(1,(char *)filename2);
+	M_free(filename2, "MultiRun Spectator filename");
 /*
 	Make sure there is space in the AM.spectatorfiles array
 */


tueda:
Indeed this is nice, though you will lose the perfect control of your spectator file names, for example, the file suffix (maybe it does matter, maybe not). The standard(?) solution is putting PID to the spectator file names, e.g., https://github.com/benruijl/forcer/blob/68a30bfb2aec3672a2f674ca8666bfc970105e24/forcer/forcer-manual.h#L131. I'm not sure if `((int)GetPID())%100000` could be always unique on a machine.

vermaseren:
Hi Josh,

If you look in the file startup.c in the routine ReserveTempFiles and then look for the variable AM.MultiRun
you will see how the -M flag is applied to the tmp files.

Cheers

Jos

jodavies:
Yes, indeed, this is where I found that `((int)GetPID())%100000` was used for this mode. I simply copied the convention also for spectator files.

By the way, is there any reason the user should have complete control over the name of spectator files, in contrast to the name of all other temp files?

vermaseren:
The reason is that you could have several spectator files at the same time. In that case you have
to know how to address them. Imagine that I were to deal with the bebe and cross topologies
in forcer simultaneaously. Either I have to put the ‘fallout’ into the same spaectator, or each in
their own (as is done now, although we do not do it simultaneously).
I just wanted to leave this possibility of having more than one.

Jos

jodavies:
Aside from the user not being able to control the file extension (if that is really important I suppose the PID could be inserted before the extension?) is there any reason not to merge this one?

Given Issue #112 , I would actually think it is not a bad idea to just force `-M` mode in all cases. Then less can go wrong for users who are not aware of potential pitfalls.

tueda:
> Aside from the user not being able to control the file extension (if that is really important I suppose the PID could be inserted before the extension?) is there any reason not to merge this one?

I feel it's OK, but not enough to push it forward for me.

I would prefer to keep file extensions, but it needs some work: what if the case like `/tmp/abc.123/spectator` or on Windows.

In the first place, the root of the problem is why `CreateSpectator` needs the filename, I think. `CreateSpectator <ExprName>` should be enough at the user level, and then FORM could automatically assign the next spectator file name on the file system.

Any input from others?

> Given Issue #112 , I would actually think it is not a bad idea to just force `-M` mode in all cases. Then less can go wrong for users who are not aware of potential pitfalls.

Maybe we need another issue focusing on the `-M` mode.

jodavies:
Right, giving FORM full control over the filenames, such that it can take its "usual precautions" also solves this issue.

tueda:
If any member with writing privileges to the repository wants to merge this PR, then I won't oppose it. It is true that the PR is one of the practical solutions that can be implemented easily and simply.

jodavies:
Any further thoughts on this one? I think relying on the user to put the PID in the spectator filename themselves in the form script when necessary is a recipe for problems in the end.

FORM5 could be the time to take user control of the spectator filenames away completely.

jodavies:
Good question. The sort files can always be called various names depending on whether form sees an "xxx" file already. I will try to test -- I have never used the checkpoint mechanism before.

PR #285, opened 2018-06-20
Change behaviour of "On fewerstatistics 0"

jodavies:
Never print statistics for small buffer sort, if ShortStatsMax is 0

Often I want statistics enabled, but only for the final sort (to see the number of terms in output).
For large calculations, one can get many thousands of lines for small-buffer sorts, even with fewerstats set as large as possible.

### DIFF
--- a/doc/manual/statements.tex
+++ b/doc/manual/statements.tex
@@ -3813,7 +3813,7 @@ the compression becomes very slow and doesn't gain very much extra.}
 small buffer is full. The keyword can be followed by a positive integer in 
 which case one out of that many of these statistics will be printed. If no 
 number is given the default value of 10 is used. When the number that 
-follows is zero, this feature is turned off (same effect as the value one).}
+follows is zero, statistics are never printed when sorting the small buffer.}
 
 \leftvitem{3.5cm}{fewerstats\index{on!fewerstats}}
 \rightvitem{13cm}{Same as the above fewerstatistics.}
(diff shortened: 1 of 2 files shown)


jodavies:
Does anyone have any opinions on this? It's a trivial change, but I am re-applying it every time I pull new commits from the repo...

tueda:
I think this is OK; hopefully, no one relies on the old behaviour of `fewerstatistics 0` (for the meaning of `fewerstatistics 1`)...

tueda:
@vermaseren 
Jos, is it OK to merge? This is backwards-incompatible, in a strict sense. Maybe you know some background of the decision of the behaviour of `fewerstatistics 0`.

vermaseren:
I guess it is OK.
Internally it affects only the printing of the statistics in the WriteStats routine as far as I know.
It may change some outputs in the sense that the number of statistics printouts changes.
This again could potentially affect scripts that process outputs.
But the value 1 is still available and it would probably be rather easy to spot this and
also very easy to repair, using grep.
Just do not forget to change the manual.

Jos

tueda:
The document change is also included in this pull request, so I will merge it...

jodavies:
It looks like the change in `sort.c` due to this was reverted in the FORM5 code merge (but the manual is intact).

tueda:
OK, I will re-apply it.

Issue #14, opened 2014-06-03
Exit may fail to exit with TFORM


tueda:
Sometimes the Exit statement causes strange things with TFORM. The following program should quit immediately,

```
S a,x1,...,x9;
L F = a*(x1+...+x9);
*L F = a+(x1+...+x9);
.sort
if (count(a,1)) exit;
.end
```

but `for i in {1..100}; do tform -w4 test.frm; done` (in bash) easily gets stuck. I also got "double free or corruption (fasttop)" errors from glibc.

Because `L F = a+(x1+...+x9)`, i.e., only one term with "a", doesn't lead this problem, I guess it comes from that two or more threads try to exit.

tueda:
Note for future development: I feel the underlying problem would be that `Terminate()` is doing many things. Actually it should be empty (except some debugging stuff). When a program exits, the OS must release all memories, close all streams and delete all temporary files (`mkstemp` + `unlink` in POSIX). Then all a worker have to do is printing an error message if needed (caring possible deadlock), maybe sending a message to the master and then ending the thread.

vermaseren:
Sounds like a good idea.
The ‘own cleanup’ stems from times that things were not that standard (at least in my mind).

Jos

tueda:
I'm not 100% sure whether this is a correct way, but commenting out `TerminateAllThreads()` in `Terminate()` helps in my environment.
https://github.com/vermaseren/form/blob/d15ec75ee60f5e0247f805f19b8d3b7613565a2b/sources/startup.c#L1740
Maybe it would be better to check whether or not the caller is a worker thread.

vsht:
If a tentative fix is available, would it be possible to propagate it into the main repository?
I would be happy to test it in my environment as well.

The problems that this bug causes are very real and annoying when you run multiple jobs on a cluster. Suppose that some of the jobs (amplitude calculations) crash because of problems with your FORM code. 

However, since `tform` doesn't exit but simply hangs with something like
`
Program terminating in thread 3 at FORM-test-script.frm Line 5
`
(this is for Takahiro's example program when run with `tform -w4`),
neither the grid control system nor the user can immediately recognize such failed jobs.
Your only safe bet it to constantly monitor the logs and grep for `terminating` or so. Then
you have to manually cancel the remaining jobs. This is not very satisfactory, since
normally the grid control system could do it for you automatically. But for that `tform` needs
to crash properly, not just hang.

vermaseren:
[...]

PR #289, opened 2018-08-05
Better error "Sorted function argument too long"

jodavies:
Print the required MaxTermSize in words to allow easy configuration of setup.

### DIFF
--- a/sources/polywrap.cc
+++ b/sources/polywrap.cc
@@ -166,7 +166,8 @@ WORD *poly_gcd(PHEAD WORD *a, WORD *b, WORD fit) {
 	if ( fit ) {
 		if ( newsize*sizeof(WORD) >= (size_t)(AM.MaxTer) ) {
 			MLOCK(ErrorMessageLock);
-			MesPrint("poly_gcd: Term too complex. Maybe increasing MaxTermSize can help");
+			MesPrint("poly_gcd: Term too complex (%d words). Maybe increasing MaxTermSize (%d words) can help",
+				newsize, AM.MaxTer/sizeof(WORD));
 			MUNLOCK(ErrorMessageLock);
 			Terminate(-1);
 		}
@@ -351,7 +352,8 @@ WORD *poly_divmod(PHEAD WORD *a, WORD *b, int divmod, WORD fit) {
 		if ( fit ) {
 			if ( ressize*sizeof(WORD) > (size_t)(AM.MaxTer) ) {
 				MLOCK(ErrorMessageLock);
-				MesPrint("poly_divmod: Term too complex. Maybe increasing MaxTermSize can help");
+				MesPrint("poly_divmod: Term too complex (%d words). Maybe increasing MaxTermSize (%d words) can help",
+					ressize, AM.MaxTer/sizeof(WORD));
 				MUNLOCK(ErrorMessageLock);
 				Terminate(-1);
 			}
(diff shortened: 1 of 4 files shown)


tueda:
Sounds nice. Possible problems would be:
1. Could it make *every* sorting in arguments a bit slow? If so, we should compute the required WORD size only when the argument doesn't fit.
2. Maybe you need to consider `ErrorMessageLock` for the change in `Malloc1()`? (I'm not sure because anyway `Error1()` causes the catastrophe #14 via `Terminate()`.)

vermaseren:
Yes, Terminate is a form of catastrophe.
At the moment I am studying what can be done when a term does not fit, such that
the job can get control back at for instance at a #enddo or something.
I am trying to deal with many cases but quite a few have terms that are too big.
I would like to be able to continue with the next case/term.
This is however far from trivial. It would have to unwind the workspace and it will
never be 100% safe.

The sorting in arguments checks usually in EndSort whether it fits. That check
gives (in case of failure) an error message and returns with an error code.
It are the calling routines that then decide to call Terminate rather than also returning
with a proper error code. It is that ‘unwinding’ that needs to be set up.
But to do that brutishly means that the program itself continues as if everything is 
fine. Hence there has to be a mechanism that restores some state (like at the start of the
module (if possible), and lets the user know somehow (per force) that something failed.
If then there is some ‘OnFailure’ handling, the program takes that, while if it is absent
the program crashes.

Maybe you have some commentary that gives a better scheme?

Jos

tueda:
Another thing: I don't see any reason why `tmpss` is declared in the beginning of `EndSort`. It can be just put in the `else` block. Physicists and programmers love the principle of locality :-)

jodavies:
I would imagine that all data is in cache, but I did not do any benchmarks. The new way avoids the problem by only checking if indeed it is too large; also then you do not need `tmpss`. Declaration wise, I was trying to follow the style of most of the form code.

In `tools.c`, there are `ErrorMessageLock` if `MALLOCDEBUG` is not defined, which is the usual state, no?

vermaseren:
MALLOCDEBUG is very special for when you are trying to see where you are writing outside
allocated memory. It can be very expensive, depending on the options you use.
In its most rigorous version it puts some extra allocations before and after the malloced
memory and fills it with zeroes. When you free the memory, it checks that there are still
only zeroes. This can make things rather slow. Hence it is off by default and only to be
used when you become desperate during debugging.
Of course nowadays there are other tools that can do similar things, and probably much
better. This MALLOCDEBUG is rather old (from the 90’s) but I figured that it is also not a
good idea to remove it.

Jos

tueda:
> In tools.c, there are ErrorMessageLock if MALLOCDEBUG is not defined, which is the usual state, no?

Ah, you are absolutely right. Somehow I confused between `#ifdef MALLOCDEBUG` and `#ifndef MALLOCDEBUG` blocks. As Jos said, it is not the place that emits errors about `MaxTermSize` and there are many places. But they are beyond of the scope of this patch, so to me it is fine to merge it.

@vermaseren and @benruijl, Opinions for Josh's patch?  (And what would be a good process to proceed and merge such PRs?)

vermaseren:
Hi,

I ran into such a thing once creating examples for my Form course.
The problem is with MaxNumberSize being normally half MaxTermSize.
If your polynomials have more normal size coefficients it pays to set
MaxNumberSize to ‘reasonable’ values like a few Kilobytes.
At least in my case that solved the problem.

Jos

vsht:
I think that a better error message here would be very useful. The current one only creates unnecessary confusion, as in #538

Issue #188, opened 2017-05-30
Relaxing the order of statements for some declarations/module options


tueda:
The fact that FORM forces users to follow the strict order of statements gives unnecessary complexity for some cases.

Consider a procedure containing executable statements with local $-variables, for example,
```
#procedure Foo()
  $FooPRIVATEx = term_;
* Use $FooPRIVATEx locally.
#endprocedure
```
which can be called like
```
if (count(SomeLabel,1));
  #call Foo()
endif;
```
A problem occurs when one wants to parallelize the program, which requires `ModuleOption local,$FooPRIVATEx` at the end of the module due to the order of the statements.

A solution for this would be to pass temporary $-variables as arguments of a procedure
```
#procedure ProcWithManyTempVars(tmp1,...,tmp9)
  `tmp1' = term_;
* ...
#endprocedure

* ...
#call ProcWithManyTempVars(<$tmp1>,...,<$tmp9>)
* ...
ModuleOption local,<$tmp1>,...,<$tmp9>;
.sort
```
Another solution that I saw is to make an additional procedure just for the module options
```
#procedure ProcWithManyTempVars()
  $procPRIVATEx = term_;
* ...
#endprocedure

#procedure ProcWithManyTempVarsModuleOptons()
  ModuleOption local,$procPRIVATEx;
* ...
#endprocedure

* ...
#call ProcWithManyTempVars()
* ...
#call ProcWithManyTempVarsModuleOptions()
.sort
```
But it would be smart to allow `ModuleOption local` to be at any places. Then we could write
```
#procedure Foo()
  $FooPRIVATEx = term_;
* Use $FooPRIVATEx locally.
  ModuleOption local,$FooPRIVATEx;  * Nice if allowed!
#endprocedure
```

A test implementation is at https://github.com/tueda/form/commit/loosedecl. This patch also relaxes the order for some of declarations (for example `Symbols`) such that the following code works
```
#procedure deriv(x)
  S derivPRIVATEy;
  id `x'^derivPRIVATEy? = derivPRIVATEy * `x'^derivPRIVATEy / `x';
#endprocedure

S y;
L F = y^3;
#call deriv(y)
#call deriv(y)
P;
.end
```

vermaseren:
Hi Takahiro,

How about if the ordering information is kind of ‘reset’ inside the procedure?
This would mean that from the viewpoint of the procedure it is a module by itself (or more than one if
[...]

PR #296, opened 2018-10-26
Relax the statement order for symbol/$-variable declarations

tueda:
From a forgotten Issue #188. We don't need to follow the FORTRAN style for symbol/$-variable declaration:
```form
CF f;
L F = f(3);
S n;
id f(n?) = n;
P;
.end
```

### DIFF
--- a/sources/compiler.c
+++ b/sources/compiler.c
@@ -43,6 +43,7 @@
 */
 
 #include "form3.h"
+#include "comtool.h"
 
 /*
 	com1commands are the commands of which only part of the word has to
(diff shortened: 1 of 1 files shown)


tueda:
Maybe I just forgot to add `noparallel`, or maybe some problem; I don't remember exactly. For consistency, other module options could be added, too.

On the other hand, if in one module both `local` and `noparallel` are specified, then it indicates some conflicts between a part that should be parallelized and a part that should be executed in the non-parallel mode. Errors/warnings could be useful in such cases.

tueda:
Maybe it would not be a problem if we give warnings/errors for inconsistent use of `ModuleOptions`, for example using both `local` and `noparallel` or `polyratfun` twice with different functions/options. The latter is dangerous because it occurs when two subroutines (procedures called in a module) assume different polyratfuns and probably leads to a wrong result.

jodavies:
What is actually inconsistent about using both `local` and `noparallel`? This should be in the end equivalent to running the code with `form` or `tform -w1` should it not?

tueda:
Yes, you are right. But this potentially indicates a performance issue (parallel part and no-parallel part in a module mistakenly), which is easily found if one has to write `ModuleOption`s only at the end of a module while may be difficult to recognize when `ModuleOption`s are distributed over a big module, especially inside procedures.

jodavies:
I think it would be interesting to resurrect this for v5?

tueda:
Actually, which ordering relaxations should we include in version 5? Do we have consensus?

tueda:
Yes. So the question is the purpose of MIXED2. Maybe it was created for these declarations and module options?

PR #394, opened 2021-09-02
Allow setting ContinuationLines to 0, meaning infinity (fixes #392)

### LINKED ISSUE #392: Removing line continuation limit
The docs describe the `ContinuationLines` option as the following:

> ContinuationLines
>
> The number of continuation lines that the local Fortran compiler will allow. This limits the number of continuation lines, when the output option `Format Fortran' (see 7.60) is selected.

This says that there's only a limit on continuation lines when a Fortran output format; this is not so: the limit is there in all formats. Here's an example with C format:

```
$ cat >test.frm <<EOF
#: ContinuationLines 1
auto symbol x;
local ex = (xabcdefg + xhijklmnop)^200;
print ex;
.sort
format C;
#write <out> "%e" ex
.end
EOF
$ form test.frm
$ cat out
pow(xhijklmnop,200) + 200*xabcdefg*pow(xhijklmnop,199) + 19900*pow(
      xabcdefg,2)*pow(xhijklmnop,198);
      _ +=  + 1313400*pow(xabcdefg,3)*pow(xhijklmnop,197) + 64684950*
      pow(xabcdefg,4)*pow(xhijklmnop,196);
      _ +=  + 2535650040*pow(xabcdefg,5)*pow(xhijklmnop,195) + 
[...]
```
Here you can see the `_ +=` marks every seconds line; those stem from the line continuation limit.

First of all, I think the documentation should be updated.

Second, would it be possible to have an option to disable this behavior? I'd like to effectively set `ContinuationLines` to infinity. In pySecDec we have a workaround that specifically recognizes the `; _ +=` sequences and replaces them with blanks, and I'd like to remove this hack completely. Unfortunately just setting `ContinuationLines` to a large number is not an option (it will only postpone the problem); it really has to be infinite.

magv:
Allow disabling the continuation line limit by setting `ContinuationLines` to 0.

Also fix the documentation: the continuation line limit applies not only for Fortran, but for all formats.

This fixes issue #392.

### DIFF
--- a/check/features.frm
+++ b/check/features.frm
@@ -811,3 +811,27 @@ assert result("F") =~ expr("
        + f(10,x3,x1,x4)
 ")
 *--#] Issue187 : 
+*--#[ Issue392_ContinuationLines_1 :
+#: ContinuationLines 1
+* Setting ContinuationLines to 0 should remove continuation line limit.
+auto symbol x;
+local ex = (xabcdefg + xhijklmnop)^100;
+.sort
+format C;
+#write <out.c> "%e" ex
+.end
+assert succeeded?
+assert file("out.c") =~ /[_] [+]=  /
+*--#] Issue392_ContinuationLines_1 :
+*--#[ Issue392_ContinuationLines_0 :
+#: ContinuationLines 0
+* Setting ContinuationLines to 0 should remove continuation line limit.
+auto symbol x;
+local ex = (xabcdefg + xhijklmnop)^100;
+.sort
+format C;
+#write <out.c> "%e" ex
+.end
+assert succeeded?
+assert !(file("out.c") =~ /[_] [+]=  /)
+*--#] Issue392_ContinuationLines_0 :
--- a/doc/manual/setup.tex
+++ b/doc/manual/setup.tex
@@ -71,9 +71,11 @@ it should not go much beyond 1000 on a 32\index{32 bits} bit machine. On a
 64\index{64 bits} bit machine it can go considerably further.}
 
 \leftvitem{4.0cm}{ContinuationLines\index{setup!continuationlines}\index{continuationlines}}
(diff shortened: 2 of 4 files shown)


magv:
Anyone? I'd imagine this change to be uncontroversial.

benruijl:
I am also for this change. In another thread @vermaseren mentioned that there may be some fixed-size buffers. Are we sure that this change does not produce buffer overflows?

magv:
There's a fixed-size buffer that holds the content of a line; this one is flushed every line. The line continuation marks are just printer every few lines or so, no buffer involved.

vermaseren:
Basically what needs to be done is rather simple:
1: whenever an output expression is printed set a flag.
2: when the flag is set: don’t print end-of-line or carriage-return.
3: when done printing the output expression, reset the flag and 
    print an end-of-line cq carriage-return.
This needs intervention in the places where you print expressions (in sch.c)
and where you print dollars. In addition you need to look for the LINEFEEDS
etc and put the if around it.

magv:
There's unfortunately also some logic to insert spaces, `&`, and `\` marks into the line buffers that would need to be modified to fully allow printing without newlines -- and that logic is a bit too tangled for me, so at the moment I've given up on changing that. Instead of the full solution, at least allowing to suppress the line continuation break marks (the `_ += ` sequences) would make handling FORM output a bit more bearable for us -- hence this pull request.

tueda (commented):
I think this PR is OK (unless it interferes with the version 5.0 development source code).
Maybe you can add a test (like `check/features.frm#Issue392`) in another commit, which guarantees this feature won't be broken.

magv:
Here's a test case.

tueda (commented):
Perfect!

benruijl:
Is this ready for merging?

magv:
If you're asking me, then yes. Not sure about others' opinions.

PR #423, opened 2022-12-02
fix some typos in sources

fchapoton:
found using the "codespell" program. This should be easy to review.


tueda:
Thanks for the pull request! 

Though I feel the change of "commutating" to "commuting" in `proces.c` is not necessary, I accepted it as we have `Commuting` and `CommuteInSet` statements in FORM.

Wikipedia has a page for ["Commutative property"](https://en.wikipedia.org/wiki/Commutative_property), so others may like "commutative".

PR #459, opened 2023-08-25
just fixing a few typos in sources/

fchapoton:
found using

`codespell -L=nd,te,nin,ba,sav,pres,numer,dum,childs sources/`

### DIFF
--- a/sources/grcc.cc
+++ b/sources/grcc.cc
@@ -55,8 +55,8 @@ using namespace Grcc;
 
 static OptDef optDef[] = {
  {"Step",       "Generate particle assigned graphs",               GRCC_AGraph, 0},
- {"Outgrf",     "Output to file (out.grf)",                               False, 0},
- {"Outgrp",     "Output to file (out.grp)",                               False, 0},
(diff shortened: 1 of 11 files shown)


sources/grcc.cc
```diff
@@ -55,8 +55,8 @@ using namespace Grcc;
 
 static OptDef optDef[] = {
  {"Step",       "Generate particle assigned graphs",               GRCC_AGraph, 0},
- {"Outgrf",     "Ouput to file (out.grf)",                               False, 0},
- {"Outgrp",     "Ouput to file (out.grp)",                               False, 0},
+ {"Outgrf",     "Output to file (out.grf)",                               False, 0},
+ {"Outgrp",     "Output to file (out.grp)",                               False, 0},
  {"1PI",        "Only 1PI graphs",                                       True , 0},
```
tueda, line 60: Sorry for nitpicking, but could you align `False, 0}` with the other lines?

tueda (changes requested):

sources/grcc.cc
fchapoton: done

tueda:
Thank you!

PR #472, opened 2024-02-22
Add brackets around dot products for format mathematica

### LINKED ISSUE #460: Improving format mathematica?
Dear FORM developers,

I guess that all FORM users exporting their results to Mathematica are aware of the issue that happens when
naively trying to save expression with scalar products, e.g.

```
Auto V p;
Auto S s;

L exp = s1*p1.p2 + s2*p3.p4^2 + s3*p3.p4^3;

.sort 
Format Mathematica;
print;
.end
```
This yields
```
 p1 . p2*s1 + p3 . p4^2*s2 + p3 . p4^3*s3
```
which Mathematica happens to interpret as 
```
Plus[Times[s1,Dot[p1,p2]],Times[s2,Dot[p3,Power[p4,2]]],Times[s3,Dot[p3,Power[p4,3]]]]
```
as compared to the correct version
```
      p1 . p2*s1 + ( p3 . p4)^2*s2 + (p3 . p4)^3*s3
```
yielding
```
Plus[Times[s1,Dot[p1,p2]],Times[s2,Power[Dot[p3,p4],2]],Times[s3,Power[Dot[p3,p4],3]]]
```
I know that one could do something like
```
repeat;
id v1?.v2? = sp(v1,v2);
id (v1?.v2?)^(-1) = pow(sp(v1,v2),-1);
endrepeat;
```
to circumvent the issue, but perhaps it could be nice if FORM could produce the right expressions right away?

Cheers,
Vlad

jodavies:
Mathematica does not interpret "a.b^2" as intended, add brackets around dot products when printing with `Format mathematica;` regardless of the presence of a power.

Fixes #460

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2297,3 +2297,47 @@ assert result("F3") =~ expr("+D")
 .end
 assert succeeded?
 *--#] Issue434 :
+*--#[ Issue460_1 :
+* Improve format mathematica, for powers of dot products
+Vector p,q;
+Local test = p.q + p.q^2 + 1/p.q + 1/p.q^2;
+Format Mathematica;
+Print;
+.end
+assert succeeded?
+assert result("test") =~ expr("(p.q)^(-2) + (p.q)^(-1) + (p.q) + (p.q)^2")
+*--#] Issue460_1 :
+*--#[ Issue460_2 :
+* Improve format mathematica, for powers of dot products
+* Check regular mode is unchanged:
+Vector p,q;
+Local test = p.q + p.q^2 + 1/p.q + 1/p.q^2;
+Print;
+.end
+assert succeeded?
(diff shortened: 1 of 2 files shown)


tueda:
The patch seems to work.

Can you push also a commit for some test case(s)? You can make, in `check/fixes.frm`,
```
*--#[ Issue460 :
* Improving format mathematical?

...

.end
assert succeeded?
assert result("...") =~ expr("...")
*--#] Issue460 : 
```
(`Issue460_1`, `Issue460_2` etc. if you need more.)

jodavies:
I wanted to also add a 4th test of the Fortran format, but could not work out how to match the output `& p_q**(-2) + p_q**(-1) + p_q + p_q**2`. Any ideas?

tueda:
OK, I see you squashed the commits. I will merge it.

PR #478, opened 2024-02-25
check.rb: terminating timed-out processes properly

tueda:
This patch makes `check.rb` properly kill test processes that have timed out.

@jodavies Could you give it a try?


jodavies:
Looks fine on my system. And the Windows CI is not taking 360 mins to fail anymore.

tueda:
Thanks. I will merge it.

> And the Windows CI is not taking 360 mins to fail anymore.

Oh, I didn't check the Deploy workflow carefully and didn't notice that the Windows jobs reached the 6-hour limit.

Issue #336, opened 2019-11-26
PolyRatFun Division by zero during normalization


jodavies:
Related: #178 , possibly #270.

There are still division by zero problems when using PolyRatFun, in both form and tform (#270 seems to be a tform-only problem).

I don't have a minimal example currently, terms which cause the problem (found using `print "%t";) do not cause problems when run alone in a simple script.

I have bisected the problem to a0b635c6ba559227abc4bd7cdbbfe272af4c2941.

The polynomials here are not large, a crash is obtained for example when the product
```
prf(1,1 + t)*prf( - 40 - 40*t - 136*t^2 - 40*t^3,9*t^2 + 18*t^3 + 9*t^4)
```
is produced, with output
```
prf( - 40 - 40*t - 136*t^2 - 40*t^3,0)
```
which causes `Division by zero during normalization` at the following sort.

There are no valgrind errors.

The DEBUG spam for this term is as follows:
```
*** [57.578]  CALL: gcd(0,9*a^3)                                   
*** [57.5781]  CALL : poly_ratfun_normalize                        
*** [57.5781]  CALL : poly_ratfun_read                             
*** [57.5781]  CALL: gcd(1,a+1)                                    
*** [57.5781]  CALL: integer_content(a+1)                          
*** [57.5781]  RES : integer_content(a+1) = 1                      
*** [57.5781]  CALL: integer_content(1)                            
*** [57.5781]  RES : integer_content(1) = 1                        
*** [57.5781]  CALL: integer_gcd(1,1)                              
*** [57.5781]  RES : integer_gcd(1,1) = 1                          
*** [57.5781]  CALL: gcd(1,a+1)                                    
*** [57.5781]  CALL: integer_content(a+1)                          
*** [57.5781]  RES : integer_content(a+1) = 1                      
*** [57.5781]  CALL: integer_content(1)                            
*** [57.5781]  RES : integer_content(1) = 1                        
*** [57.5781]  CALL: integer_gcd(1,1)                              
*** [57.5781]  RES : integer_gcd(1,1) = 1                          
*** [57.5781]  CALL: gcd(1,1)                                      
*** [57.5781]  CALL : poly_ratfun_read                             
*** [57.5781]  CALL: gcd(1,9*a^2+18*a^3+9*a^4)                     
*** [57.5781]  CALL: integer_content(9*a^2+18*a^3+9*a^4)           
*** [57.5781]  RES : integer_content(9*a^2+18*a^3+9*a^4) = 9       
[...]

Issue #270, opened 2018-02-19
Problems with tform and polyratfun


jodavies:
Hello,

I have not (yet?) found a particularly minimal example for this issue, but I will post what I have so far.

With the following code, I am finding that in some* runs, the final result (which should be 0) is
```
       + ep^4*prf(m4^7 - m4^7,m4^14 + 2*m4^7 + 1)
```
or
```
       + ep^4*prf(m4^7 - m4^7,m4^7 + m4^14 + 1 + m4^7)
```

The problem occurs in both tform and tvorm, 4,1, 4.2.0 and the latest git, and tvorm inside valgrind (but valgrind does not report any errors).

*: problems start at -w3.
For 3,5 threads there are problems ~ 2.4%,1.4% of the time. -w3 seems to produce *only* the first output above. All other cases produce both.
For even numbers of threads the problem is less frequent, more like 0.13% of runs.

The code is as follows. I have attached the necessary includes. They need renaming to .h.
```
#-
Off Statistics;

CFunction prf,den;
AutoDeclare Symbol n,m,ep;

#include- test-expanded-tiny.h
.sort
PolyRatFun prf;

#message Loading mma results...
#include- test-result-tiny.h
.sort

Drop;
Local diff7 = test7 - result7;
.sort

Denominators den;
FactArg den;
Repeat Identify den(?a,ep,?b) = 1/ep*den(?a,?b);
Repeat Identify den(n?,m?,?a) = den(n*m,?a);
Identify den(n?) = prf(1,n);

#do i = 1,5
	Identify m`i'^n? = prf(m`i'^n,1);
	Identify n`i'^m? = prf(n`i'^m,1);
#enddo
.sort

Print +s diff7;

.end
```
[test-expanded-tiny.txt](https://github.com/vermaseren/form/files/1736643/test-expanded-tiny.txt)
[test-result-tiny.txt](https://github.com/vermaseren/form/files/1736644/test-result-tiny.txt)

I will attempt to find a more trivial example later...

Thanks,
Josh.

tueda:
An experiment (though I don't know the meaning of this probablity distribution and also fluctuation uncertainty):
```
-w3 count=10000 good=9543 bad=457
-w4 count=10000 good=9869 bad=131
-w5 count=10000 good=9750 bad=250
-w6 count=10000 good=9578 bad=422
-w7 count=10000 good=9761 bad=239
-w8 count=10000 good=9815 bad=185
```
*to be updated...*

[...]

PR #482, opened 2024-02-26
Fix polyratfun bugs caused by incorrect clean flag

jodavies:
There are some known issues regarding polyratfun which are not easy to fix.

The least we can do is to crash when we hit something that is definitely wrong, for now.

This "fixes" the example in Issue #336 

This does not catch Issue #270 !

### DIFF
--- a/sources/poly.cc
+++ b/sources/poly.cc
@@ -2560,12 +2560,24 @@ const poly poly::argument_to_poly (PHEAD WORD *e, bool with_arghead, bool sort_u
 	}
 
 	res[0] = ri;
-	
-	// normalize, since the Form order is probably not the polynomial order
-	// for multiple variables
 
-	if (sort_univar || AN.poly_num_vars>1)
+	// JD: For univariate cases, check whether the ordering is correct or not.
+	// There are various ways to arrive here, but with an incorrect ordering, such
+	// as interactions with collect, moving a polyratfun out of another function
+	// argument, etc.
+	// If the ordering is not correct, make sure we normalize. In multivariate cases,
+	// always normalize as before.
+	if (sort_univar == false && AN.poly_num_vars == 1) {
+		if (res.number_of_terms() >= 2) {
+			if(res[2] < res.last_monomial()[2]) {
+				sort_univar = true;
+			}
+		}
+	}
+
+	if (sort_univar || AN.poly_num_vars>1) {
 		res.normalize();
+	}
 
 	NumberFree(den,"poly::argument_to_poly");
 	NumberFree(pro,"poly::argument_to_poly");
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2173,6 +2173,122 @@ print +s;
 assert succeeded?
 assert result("A") =~ expr("+N1")
 *--#] Issue325 :
+*--#[ Issue336_1 :
+#-
+CFunction rat,f,tag;
+Symbol a,b,y,x;
+
+* Build lots of combinations
+Local test = (
+	+ f(1)
+	+ f(1+x)
+	+ f(1+x+x^2)
+	+ f(x+x^2)
+	+ f(1+x^-1)
+	+ f(1+x^-2)
+	+ f(1+x+x^-1)
+	+ f(x+x^-1)
+	+ f(x^-1)
+	+ f(x^-2)
+	+ f(x^-1+x^-2)
+	+ f(1+x+y)
+	+ f(1+x+x^2+y+y^2)
+	+ f(x+x^2+y+y^2)
+	+ f(1+x^-1+y^-1)
+	+ f(1+x^-2+y^-2)
+	+ f(1+x+x^-1+y+y^-1)
+	+ f(x+x^-1+y+y^-1)
+	+ f(x^-1+y^-1)
+	+ f(x^-2+y^-2)
+	+ f(x^-1+x^-2+y^-1+y^-2)
+	+ f(x^-1*x^-1)
+	+ f(x^-2+x^-1*y^-1+y^-2)
+	+ f(x^2+x^1*y^-1+y^-2)
+	)^2;
+DropCoefficient;
+* Create rational polys, with both orderings of the num and den
+Identify f(a?)*f(b?) = rat(a,b) + rat(a,b);
+.sort
+DropCoefficient;
+#$i = 0;
+Multiply tag($i);
+$i = $i+1;
+Print +s;
+.sort
+
+* Everything should cancel in the end, and we should get zero.
+Identify rat(?a) = rat(?a) - x*rat(?a);
+Bracket x;
+.sort
+
+PolyRatFun rat;
+Collect f;
+.sort
+Identify x = 1;
+Identify f(x?) = x;
+
+Print;
+.end
+assert succeeded?
+assert result("test") =~ expr("0")
+*--#] Issue336_1 :
+*--#[ Issue336_2 :
+#-
+CFunction rat,f,tag;
+Symbol a,b,y,x;
+
+* Build lots of combinations
+Local test = (
+	+ f(1)
+	+ f(1+x)
+	+ f(1+x+x^2)
+	+ f(x+x^2)
+	+ f(1+x^-1)
+	+ f(1+x^-2)
+	+ f(1+x+x^-1)
+	+ f(x+x^-1)
+	+ f(x^-1)
+	+ f(x^-2)
+	+ f(x^-1+x^-2)
+	+ f(1+x+y)
+	+ f(1+x+x^2+y+y^2)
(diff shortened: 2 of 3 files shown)


jodavies:
Looking through the issues, I think that the known open polyratfun bugs are now caught and terminate form, so the user can adjust their script to avoid them, though of course they are not fixed. Without looking at the complicated interactions of polyratfun with collect, argument etc in more detail, this seems like a good intermediate change to make, for safety. What do you think?

jodavies:
At least with @tueda 's benchmark ( https://gist.github.com/tueda/8cabf511573b115b9c17a7a181bf0248 ) there is no measurable performance difference due to this.

tueda:
The code compares the powers of the first variables in the first and last terms. It is true that at least it catches problems in particular cases, e.g.,
```form
CF rat,rat1,f;
S x,y;
L F = rat(1,1+x)*rat1(1,1+x^2);
id rat1(?a) = f(rat(?a));
.sort

PolyRatFun rat;
print "1: %t";
id f(x?) = x;
print "2: %t";
.end
```
But I'm not sure what really causes the error in the following sense: I can't make a simple case that is wrong but can't be caught by this check. For example, if I give `F = rat(1+x,1+y)*rat1(1+x^2,1+y^2);` then the order in `f(rat(...))` is wrong but somehow the check didn't hit anything (maybe `poly_ratfun_read` works well for multivariate cases??) and indeed the result is correct:
```
1:  + rat(x + 1,y + 1)*f(rat(1 + x^2,1 + y^2))
2:  + rat(x^3 + x^2 + x + 1,y^3 + y^2 + y + 1)
```
If we can find more precise conditions under which errors occur, the check may become simpler and lighter.

jodavies:
Interesting. The second variable doesn't even need to go in rat1. `rat(y,1+x) * rat1(1,1+x^2)` has no issues apparently.

This should be causing `poly_num_vars` to be 2, and so https://github.com/vermaseren/form/blob/b1f9041b7bb6f5dd4a149c87e70922ad9914b291/sources/poly.cc#L2567-L2568 always normalizes the poly regardless of whether the arg is marked clean or not in `poly_ratfun_read`.

So actually it seems like the issues can all be fixed properly, and the checks removed, by unconditionally calling `normalize()` here also in the univariate case? This will slow things down in univariate cases (by an amount to be determined in benchmarks...) but actually it will not affect multivariate polyratfun performance?

jodavies:
For @tueda 's benchmark (with default N=30), on my system I get `17.90+-0.17s` -> `18.03+-0.16s` by unconditionally calling `normalize()`.

tueda:
How about some simple check like
```
if (is_univariate)
  if (number_of_terms >= 2)
    if (power_of_variable_in_1st_term < power_of_variable_in_2nd_term)
      sort_univar = true;
```

vermaseren:
This needs some gymnastics with dollar variables, but it can be done.

tueda:
I meant adding a simple check that is negligible in time for Josh's unconditional calling `normalize()` in `poly.cc`.

vermaseren:
Sorry about that then.

jodavies:
I'll look at this and check performance. It is not quite so simple, checking powers of first and second terms fails in the case of `rat1(1,1+x^2+x^-1)`.

jodavies:
Here is a new try. Checking first and last monomials should work also for cases where symbols appear with negative exponent.

sources/poly.cc
```diff
@@ -2561,11 +2561,23 @@ const poly poly::argument_to_poly (PHEAD WORD *e, bool with_arghead, bool sort_u
...
 	
-	// normalize, since the Form order is probably not the polynomial order
-	// for multiple variables
+	// JD: For univariate cases, check whether the ordering is correct or not.
+	// There are various ways to arrive here, but with an incorrect ordering, such
+	// as interactions with collect, moving a polyratfun out of another function
+	// argument, etc.
+	// If the ordering is not correct, make sure we normalize. In multivariate cases,
+	// always normalize as before.
+	if (sort_univar == false && AN.poly_num_vars == 1) {
```
tueda, line 2570: It's a matter of taste, but `!sort_univar` is more natural than `sort_univar == false`?

Or, these lines may be written as
```c
    if (sort_univar || AN.poly_num_vars > 1 ||
        AN.poly_num_vars == 1 && res.number_of_terms() >= 2 &&
            res[2] < res.last_monomial()[2]) {
        res.normalize();
    }
```
I'm not sure whether `AN.poly_num_vars == 1` is required or not. `AN.poly_num_vars` can be 0?

tueda:
It gives the correct answer for `rat(1,1+x^2+x^-1)*rat1(1,1+x^2+x^-1)` but not for `rat(1,1+x)*rat1(1,1+x^2+1/x)`. (Is it an expected limitation of your `poly.cc` code?)

```form
CF rat,rat1,rat2,f1,f2;
S x,y,n;

L F = rat(1,1+x)*rat1(1,1+x^2+1/x);

L G = F * replace_(rat1,rat2);
id rat1?{rat1,rat2}[n](?a) = {f1,f2}[n](rat(?a));
.sort

PolyRatFun rat;
P "1: %t";
argument f2;
  argument rat;
  endargument;
endargument;
P "2: %t";
id f1?{f1,f2}(x?) = x;
P "3: %t";

.sort
Skip;
L ZERO = F - G;
P;
.end
```
```
FORM 5.0.0-beta.1 (Mar  1 2024, v5.0.0-beta.1-28-ge58eac0)  Run: Wed Mar  6 14:03:01 2024
...
1:  + rat(1,x + 1)*f1(rat(1,1 + x^-1 + x^2))
2:  + rat(1,x + 1)*f1(rat(1,1 + x^-1 + x^2))
3:  + rat(1,x^3 + x^2 + x + 2 + 1)
...
1:  + rat(1,x + 1)*f2(rat(1,1 + x^-1 + x^2))
2:  + rat(1,x + 1)*f2(rat(x,x^3 + x + 1))
3:  + rat(x,x^4 + x^3 + x^2 + 2*x + 1)
...
   ZERO =
      rat( - x + 1,x^7 + 2*x^6 + 3*x^5 + 7*x^4 + 7*x^3 + 6*x^2 + 7*x + 3);
```

jodavies:
No it's not expected, good catch. I'll take a look.

jodavies:
I am not sure what to make of this one. In the `rat(1,1+x+x^-1) * rat1(1,1+x+x^-1)` case, the rat denominator is passed to `poly::argument_to_poly` as `13 4 1 1 1 4 -1 1 1 4 0 1 1` with `sort_univar=true`, and normalize produces `13 4 1 1 1 4 0 1 1 4 -1 1 1`. The rat1 denominator is passed in as `13 4 0 1 1 4 -1 1 1 4 1 1 1` (I am not sure why this is different) with `sort_univar=false` but the new code sets it to true and we normalize again to `13 4 1 1 1 4 0 1 1 4 -1 1 1`.

In the `rat(1,1+x) * rat1(1,1+x+x^-1)` case, the denominator of rat1 is passed just as before and, with the new code, normalized as before. And yet the result is not correct. The term print there looks like `2:  + rat(1,x^2 + 2*x + 2 + 1)`.

jodavies:
OK, now I think I follow it. `poly_ratfun_read` in `polywrap.cc` also needs to know whether normalization is required. This is where the "denominator of the denominator" is multiplied into the numerator and vice versa. So the check for bad sort ordering needs to move here instead, before calling `argument_to_poly`. If I force `clean=false` there, the new "bad examples" are also correct.

jodavies:
This was quite easy to fix actually; just check for negative powers in the normalized num and den unconditionally, and then set the clean flag if they appear.

For the checks in `poly.cc`, personally I like it better as separate code since you can see the change compared to the old code better. But if you strongly prefer it all in the one if statement I can do that.

I am fairly sure we now catch the bad cases here. If you're happy I will squash all of these commits.

tueda:
> For the checks in `poly.cc`, personally I like it better as separate code since you can see the change compared to the old code better. But if you strongly prefer it all in the one if statement I can do that.

OK. That's fine.

> This was quite easy to fix actually; just check for negative powers in the normalized num and den unconditionally, and then set the clean flag if they appear.

The check is now performed with or without the `clean` flag. No performance loss?

jodavies:
Presumably there is a very small loss, I will try to measure. I think you do need to check all of the powers though, since in multivariate cases you can't just check the last term of the polynomial, since the negative power might be in a variable which is sorted earlier, and appear in the middle of the terms.

jodavies:
Do you feel there is anything further to add to this one? I think it is in a pretty good shape now.

tueda:
Unless there is some performance loss for other use cases, I think this is fine. It should be merged before the next (beta?) version release.

PR #483, opened 2024-02-27
Don't run moebius_3 test when available memory is limited

tueda:
Here, I simply set constant memory requirements. (Maybe this is enough?) In principle, one can consider the number of cpus etc. but it may be too complicated.

The serial version requires 8 GB (in SI, not GiB), so 8 GiB system memory is enough to run it. The threaded/mpi versions need 20GB. Note that [16 GiB is available on GitHub-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

### DIFF
--- a/check/check.rb
+++ b/check/check.rb
@@ -70,6 +70,19 @@ def read_env_positive_int(key, default_value)
   value
 end
 
+# Get the total size of the physical memory available on the host machine.
+def get_total_physical_memory
+  if RUBY_PLATFORM.downcase.include?("linux")
+    mem_info = `free -b | grep Mem`
+    mem_info.split[1].to_i
+  elsif RUBY_PLATFORM.downcase.include?("darwin")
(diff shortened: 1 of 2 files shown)


jodavies:
This seems good to merge to me.

Korg666999:
1. [ ] ]|[]|[][]|[]|[]|[][]|[]|[]|[][]|||::}}}}::..
2. [ ] ]|[]|[][]|[]|[]|[][]|[]|[]|[][]||||::}}}}::\::....
3. [ ] ]|[]|[][]|[]|[]|[][]|[]|[]|[][]|||||::}}}}::::....
5. [ ] ]|[]|[][]|[]|[]|[][]|[]|[]|[][]||||||::}}}}:::\::....

PR #491, opened 2024-03-07
Format mathematica: enclose expressions in brackets

### LINKED ISSUE #23: Incompatible Mathematica output
When printing expressions in `Mathematica` format sometimes output contains newlines (see example below) which are treated by Mathematica as end of the expression. Because of that Mathematica ignores everything after the first newline.

My solution is to wrap a whole output expression in parentheses, e.g. `F = ( <expr> )`, but you of course may have other ideas.

```
CF f,g;
S x;

L F = (x+1)^2*(f(x)+g(x))^5;

Format Mathematica;

Bracket x;
Print +s;
.end
```

jodavies:
If there are blank lines in the output (for eg when printing a bracketed expression) mathematica considers the expression to have finished. This is avoided by wrapping the entire expression in brackets.

Closes #23

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -186,6 +186,57 @@ P;
 assert succeeded?
 assert result("F") =~ expr("1")
 *--#] Issue21 : 
+*--#[ Issue23 :
+#-
+CFunction f,g;
+Symbol x;
+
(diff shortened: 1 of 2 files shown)


tueda:
OK. Thanks!

PR #492, opened 2024-03-11
Float test from Madrid workshop/manual

jodavies:
Here I have added the example from the Madrid workshop slides.


jodavies:
To do: why does `F1` print with `e+0` and `F2` without? Which is the desired behaviour?

tueda:
Well, if we put it into `example.frm` (examples in the manual), then should we put this example into the manual with the following comment?
```tex
% THIS EXAMPLE IS PART OF THE TESTSUITE. CHANGES HERE SHOULD BE APPLIED THERE AS
% WELL! (Float_1)
```

jodavies:
The `e+0` is a bug actually: F1 should be `e+3` (easy to see from the rational form of F1/F2).

vermaseren:
Very funny indeed. It is the term
+89/18*mzv_(9,3,3)
that is the culprit. If you leave it out, it prints without the e+0 and if you include it but leave
the last term out, it just prints e.
Gotta have a good look on the linux computers, because gdb does not work on my apple.
I can leave some more terms out to just get the e, but I have not yet discovered the ‘pattern'

jodavies:
The ouput is truncated here:
https://github.com/vermaseren/form/blob/e99543ad939ef823f5e8db0678aa18028e7de707/sources/float.c#L950

If you ask for `numdigits+1` then you get `e+03` (and now a bare `e` appears on the output of `F2`). I think the required buffer size for the print can be found by calling `gmp_snprintf(NULL,0,"%.*Fe",numdigits,aux4)` first.

jodavies:
I was too hasty there, the issue is in the post-processing code that follows. As I understand it you would like the following:
1. truncate trailing zeros
2. don't print "e+00"

Is this correct?

fllor:
I found the bug printing routine. One has to change

https://github.com/vermaseren/form/blob/e99543ad939ef823f5e8db0678aa18028e7de707/sources/float.c#L962-L963
to
```c
while ( n2 < n ) { *s1++ = *s2++; n2++; }
n -= (n2-n1);
```

This will correctly truncate all trailing zeros, but leave the exponent intact. Removing "e+00" is not part of this routine.

jodavies:
OK, please add the fix in your branch and lets see if the valgrind errors are gone. I can rebase this on top of your changes later.

tueda:
By the way, why is the precision of printed numbers strongly influenced by the maximum weight?
```form
#StartFloat 50,6
L F = 1/3;
tofloat;
P;
.end
```
```
   F =
      3.3e-01;
```

jodavies:
The final digit of pi is also not rounded correctly, is it understood why this is?

jodavies:
This is updated for the changes of #676, which implies slightly different numbers in the manual also. Since it is in the manual it should probably indeed go into the test suite, even if @cbmarini is working on other float-related testing.

PR #493, opened 2024-03-14
Move the LogType check inside WriteStats

### LINKED ISSUE #470: Log file writing failure with threadstats
When I run Tform on my Windows laptop using WSL, the log file suddenly stops being updated at some point during the calculation, at a point when the workers are running. This has happened for me on all of the 4 occasions when Threadstats was turned on, but not otherwise. The computation itself continues without a problem.

My guess is that there is a problem with the way WSL handles the the race conditions when several threads are trying to write to the file at the same time.

jodavies:
This allows other code to be included in the lock, in particular to avoid race conditions when temporarity setting `AC.LogHandle = -1` when AM.LogType is 1.

Fixes #470

### DIFF
--- a/sources/ftypes.h
+++ b/sources/ftypes.h
@@ -323,6 +323,18 @@ typedef int (*TFUN1)();
 #define SORTPOWERFIRST 2
 #define SORTANTIPOWER 3
 
+/* These control the output of WriteStats. The different sort
+ * types have differently formatted stats. These variables should
+ * not have arbitrary values since they are also used as array
+ * indices by WriteStats. */
+#define STATSSPLITMERGE 0
+#define STATSMERGETOFILE 1
+#define STATSPOSTSORT 2
+/* CHECKLOGTYPE implies that statistics will not be printed in
+ * the log file if AM.LogType = 1 (when running with -ll or -F). */
+#define NOCHECKLOGTYPE 0
+#define CHECKLOGTYPE 1
+
 #define NMIN4SHIFT 4
 /*
 	The next are the main codes.
--- a/sources/sort.c
+++ b/sources/sort.c
@@ -71,7 +71,7 @@ LONG numcompares;
 /*
   	#] Includes : 
 	#[ SortUtilities :
- 		#[ WriteStats :				VOID WriteStats(lspace,par)
+		#[ WriteStats :				VOID WriteStats(lspace,par,checkLogType)
 */
  
 char *toterms[] = { "   ", " >>", "-->" };
@@ -81,16 +81,18 @@ char *toterms[] = { "   ", " >>", "-->" };
  *
  *		@param plspace The size in bytes currently occupied
  *		@param par
- *		par = 0 after a splitmerge.
- *		par = 1 after merge to sortfile.
- *		par = 2 after the sort
+ *		par = 0 = STATSSPLITMERGE after a splitmerge.
+ *		par = 1 = STATSMERGETOFILE after merge to sortfile.
+ *		par = 2 = STATSPOSTSORT after the sort
+ *		@param checkLogType == CHECKLOGTYPE: The output should not
+ *		go to the log file if AM.LogType is 1.
  *
  *		current expression is to be found in AR.CurExpr.
  *		terms are in S->TermsLeft.
  *		S->GenTerms.
  */
 
-VOID WriteStats(POSITION *plspace, WORD par)
+VOID WriteStats(POSITION *plspace, WORD par, WORD checkLogType)
 {
 	GETIDENTITY
 	LONG millitime, y = 0x7FFFFFFFL >> 1;
@@ -107,7 +109,7 @@ VOID WriteStats(POSITION *plspace, WORD par)
 #endif
 		if ( Expressions == 0 ) return;
 
-		if ( par == 0 ) {
+		if ( par == STATSSPLITMERGE ) {
 			if ( AC.ShortStatsMax == 0 ) return;
 			AR.ShortSortCount++;
 			if ( AR.ShortSortCount < AC.ShortStatsMax ) return;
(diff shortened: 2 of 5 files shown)


jodavies:
Given the large output of helgrind and drd in https://github.com/vermaseren/form/issues/470#issuecomment-1996014250 it might be an idea to make a careful study of potential problems, but changes due to that probably belong in a separate PR.

tueda:
> it might be an idea to make a careful study of potential problems

Indeed, I was trying to resolve [Helgrind](https://valgrind.org/docs/manual/hg-manual.html) errors, [here](https://github.com/tueda/form/commits/helgrind-clean-20240313/), still have a long way to go. This is very tricky even for a simple program like
```form
L F = 1;
.end
```
For example, [this commit](https://github.com/tueda/form/commit/087b316501e557d0bf082f1f88ddb5130a6aa66d) fixes a Helgrind error by introducing a barrier, but it leads to another [drd](https://valgrind.org/docs/manual/drd-manual.html) error about barrier destruction (which must be fixed).

tueda:
I think it's OK to merge, but I have 2 concerns:
- Possible performance loss (negligible?): `WriteStats()` has many early returns, where nothing is written, so the lock is unnecessary for these cases.
- Do we need to use the lock for `WriteStats()` in the master thread (2 places in `threads.c`)? Is it logically fine without the lock after considering interferences with other locks and the program flow (too complicated to follow), or might it be OK in practice (though logically bad)?

I'm not sure but maybe the juggling of `AC.LogHandle` could be moved into `WriteStats()` at the cost of adding another argument (?)

jodavies:
With typical buffer sizes, printing of stats is supposed to be very rare compared to actual algebra processing. I would say that if these locks become a bottleneck, far more performance improvement is available to the user by adjusting buffer sizes, than by locking a bit less (after the early returns, for eg). But indeed, if an argument is added, the lock can go back inside `WriteStats`. This is quite an easy change, `WriteStats` is not called very many times across the codebase.

For your second point, I don't know. I just made sure that if there were locks before this change, there are also locks after.

jodavies:
Your suggestion is implemented in [96f040d](https://github.com/vermaseren/form/pull/493/commits/96f040d38a786d69cba70b79f0438ec7174fd725).

This is much cleaner I think.

A possible additional cleanup: make defines for par={0,1,2} for the second argument of WriteStats ?

sources/ftypes.h
```diff
@@ -311,6 +311,9 @@ typedef int (*TFUN1)();
 #define NORMALFORMAT 0
 #define NOSPACEFORMAT 1
 
+#define NOCHECKLOGTYPE 0
```
tueda, line 314: I think it's good to add a comment saying they are for `checkLogType` in `WriteStats()`.

tueda:
> This is much cleaner I think.

Very nice, thanks!

> A possible additional cleanup: make defines for par={0,1,2} for the second argument of WriteStats ?

I think this is a good idea for readability.

sources/ftypes.h
```diff
@@ -323,6 +323,18 @@ typedef int (*TFUN1)();
 #define SORTPOWERFIRST 2
 #define SORTANTIPOWER 3
 
+/* These control the output of WriteStats. CHECKLOGTYPE implies
```
tueda, line 326: It is more natural to have constants first for `par` and then for `checkLogType`.

jodavies: Thanks for the review. This is resolved in [c7b0a51](https://github.com/vermaseren/form/pull/493/commits/c7b0a51356a81b87180d945ffae3f07c720be7e6)

jodavies:
I think now that your comment regarding performance of the original commit was important. Since in WriteStats everything is inside `if ( AT.SS == AT.S0 && AC.StatsFlag ) {`, the original commit introduced unnecessary locking for all of the higher level sorts (of function arguments, etc) which never print any stats.

I'll squash these, if you are happy with the current state.

sources/sort.c
```diff
@@ -124,7 +124,7 @@ VOID WriteStats(POSITION *plspace, WORD par, WORD checkLogType)
 		 * LogHandle. This must be done within the ErrorMessageLock region to
 		 * avoid a data race between threads. */
 		const WORD oldLogHandle = AC.LogHandle;
-		if ( checkLogType && AM.LogType ) {
+		if ( checkLogType == CHECKLOGTYPE && AM.LogType ) {
```
tueda, line 127: Well, I understand why you added `== CHECKLOGTYPE`...; it's logically the right way, but typically, this generates a longer binary code because a register has to be compared with `1` (though it should be completely negligible, of course). 

([Compiler Explorer example](https://gcc.godbolt.org/z/d9TvcdYon): function `foo1` has `cmp edi,0x1` (`83 ff 01`) while function `foo2` has `test edi,edi` (`85 ff`))

At this point, we don't need to add `== CHECKLOGTYPE`, I think. If later we need to extend `checkLogType` to ternary or more, we can change it at that time.

jodavies: OK, I will drop this again in the squash.

tueda: Can you also resolve the conflict with https://github.com/vermaseren/form/commit/3ccda3a3145036a7a8def3ae3bf7b3dbe8066646?

jodavies: Done in principle... I actually don't understand why github considers what remains, a conflict.

tueda: In this case, I think you need to rebase the upstream master branch (`vermaseren/master`):
```bash
git switch issue-470
git fetch upstream  # I assume it points to vermaseren/master
git reset --hard upstream/master
git cherry-pick 0b0443e
# here the conflict should be manually resolved
git add -u
git cherry-pick --continue
```
Then
```bash
git range-diff HEAD^..HEAD 0b0443e^..0b0443e
```
should print only the resolved conflict.

If you make a mistake or have a trouble in these steps, then you can always reset the branch to `upstream/master` or `0b0443e`.

tueda:
Merged. Thanks!

Issue #334, opened 2019-11-15
Maximal length of names of expressions?


vsht:
In the manual it is written that

> 
> The length of names is not restricted in FORM. There is one exception to this rule: names of
> expressions cannot be longer than 16 characters. Of course in practise there are physical limits on
> the size of names, posed by the size of the memory of the computer being used.

However, one can obviously do something like

```
off statistics;
S s, x, y;
CF acc;
 Global thisexpressionis32characterslong = x^2+y^2;
.store

save thisexpressionis32characterslong.res thisexpressionis32characterslong;

.end
```

without triggering any warning or errors.

When trying to load the saved expression 

```
load thisexpressionis32characterslong.res;
.sort
L exp = thisexpressionis32characterslong;
.sort
print exp;
.end
```
it seems that the expression was not saved properly. However, the truth is
that FORM cuts the name after the 20th character, so

```
load thisexpressionis32characterslong.res;
.sort
L exp = thisexpressionis32ch;
.sort
print exp;
.end
```

works and recovers the correct expression.

Are things expected to work this way? Naively I would rather expect an error
message once the name length exceeds 16 characters.

Cheers,
Vladyslav

tueda:
Please see #192. The problem is `MAXENAME` for .sav files.

vermaseren:
Let us say that this is something that suffered from much tampering in the history of Form.
But then tampering by more than one person. Clearly not in a completely coherent fashion.
But if you hold to the 16 characters rule, you will always be safe.

The last tampering was years ago when the sav files were made systems independent.
Probably that stretched things from 16 to effectively 20 although, it may well be that the
excess 4 bytes are actually supposed to be used for something different if a certain flag is set,
or maybe it is just padding for now.

Jos

vsht:
I see, thanks! Would it be perhaps possible to add a flag (or something on the form.set level) that strictly enforces the 16 chars rule, so that one can actively avoid such situations?

When using automated setups with lots of scripts to generate expression names on the fly,
it is not always easy to ensure that nothing is longer than 16 chars.

Issue #192, opened 2017-06-07
Length of expression names


jodavies:
Hello,
I have just noticed that in the manual it says
```
The length of names is not restricted in FORM. There is one exception to this rule: names of expressions cannot be longer than 16 characters.
```
I regularly use Local expressions with names longer than this, and have never had any issues.

What, precisely, are the restrictions here, if any?

Thanks,
Josh.

tueda:
I digged the forum and found https://www.nikhef.nl/~form/forum/viewtopic.php?f=4&t=190&p=568.

Expressions with names more than 16 characters can't fit in the normal statistics, which may breaks programs parsing the FORM output or log (like formprof.py).
```
L F234567890123456 = 1;
L F2345678901234567 = 1;
L F23456789012345678 = 1;
L F23456789012345678901234567890 = 1;
P;
.end
```
```
FORM 4.1 (Jun  6 2017, v4.1-20131025-347-g76561be) 64-bits  Run: Wed Jun  7 16:39:17 2017
    L F234567890123456 = 1;
    L F2345678901234567 = 1;
    L F23456789012345678 = 1;
    L F23456789012345678901234567890 = 1;
    P;
    .end

Time =       0.00 sec    Generated terms =          1
F234567890123456         Terms in output =          1
                         Bytes used      =         20

Time =       0.00 sec    Generated terms =          1
2345678901234567         Terms in output =          1
                         Bytes used      =         20

Time =       0.00 sec    Generated terms =          1
3456789012345678         Terms in output =          1
                         Bytes used      =         20

Time =       0.00 sec    Generated terms =          1
5678901234567890         Terms in output =          1
                         Bytes used      =         20

   F234567890123456 =
      1;

   F2345678901234567 =
      1;

   F23456789012345678 =
      1;

   F23456789012345678901234567890 =
      1;
```
For now I don't see any other issues. The last one in the above example (30 characters) looks OK in the Print output.

vermaseren:
There is another issue here.
In the .sav files there is a fixed size for each entry.
There the maximum size has been set to MAXENAME which is 16 (see the struct of type INDEXENTRY.
If this is to be changed, .sav files are not backwards compatible. People will complain.

Jos

PR #500, opened 2024-04-26
Print a warning when saving expr with too-long names

jodavies:
This addresses #192 and #334 .

Terminating on this condition will certainly break people's existing FORM scripts.

At least print a warning, since one can certainly produce bugs by using names over MAXENAME (16) characters.

An example bug:
```
#-
Off statistics;

Global abcdefghijklmnop = 16;
Global abcdefghijklmnopq = 17;
Global abcdefghijklmnopqr = 18;
Global abcdefghijklmnopqrs = 19;
Global abcdefghijklmnopqrst = 20;
Global abcdefghijklmnopqrstu = 21;

Print +s;
.store

*Save test.res, abcdefghijklmnop, abcdefghijklmnopq;
Save test.res;
.end
```
then
```
#-
Off Statistics;

Load test.res;

Local Abcdefghijklmnop      = abcdefghijklmnop;
Local Abcdefghijklmnopq     = abcdefghijklmnopq;
Local Abcdefghijklmnopqr    = abcdefghijklmnopqr;
Local Abcdefghijklmnopqrs   = abcdefghijklmnopqrs;
Local Abcdefghijklmnopqrst  = abcdefghijklmnopqrst;
*Local Abcdefghijklmnopqrstu = abcdefghijklmnopqrstu;

Print +s;
.end
```
which gives:
```
FORM 4.1 (Apr 26 2024, v4.1-20131025-620-ga113449-dirty)  Run: Fri Apr 26 15:09:02 2024
    #-
 abcdefghijklmnop loaded
 abcdefghijklmnopq loaded
 abcdefghijklmnopqr loaded
 abcdefghijklmnopqrs loaded
 abcdefghijklmnopqrstx loaded
 abcdefghijklmnopqrst loaded

   Abcdefghijklmnop =
       + 16
      ;

   Abcdefghijklmnopq =
       + 17
      ;

   Abcdefghijklmnopqr =
       + 18
      ;

   Abcdefghijklmnopqrs =
       + 19
      ;

   Abcdefghijklmnopqrst =
       + 21
      ;

  0.00 sec out of 0.00 sec
```
Here, `abcdefghijklmnopqrst` is loaded with the incorrect name `abcdefghijklmnopqrstx`, and `abcdefghijklmnopqrstu` is loaded as `abcdefghijklmnopqrst`. Then `Abcdefghijklmnopqrst` has value `21` and not `20`.

With the patch, upon saving, you get:
```
test.frm Line 15 --> Warning: saved expression name over 16 char: abcdefghijklm
nopq
test.frm Line 15 --> Warning: saved expression name over 16 char: abcdefghijklm
nopqr
test.frm Line 15 --> Warning: saved expression name over 16 char: abcdefghijklm
nopqrs
test.frm Line 15 --> Warning: saved expression name over 16 char: abcdefghijklm
nopqrst
test.frm Line 15 --> Warning: saved expression name over 16 char: abcdefghijklm
nopqrstu
```

What do you think? I used MesPrint and not Warning/HighWarning since those don't take arguments for the format string.

### DIFF
--- a/sources/store.c
+++ b/sources/store.c
@@ -366,6 +366,7 @@ int CoSave(UBYTE *inp)
 	UBYTE *p, c;
 	WORD n = 0, i;
 	WORD error = 0, type, number;
+	WORD exprInStorageFlag = 0;
 	LONG RetCode = 0, wSize;
 	EXPRESSIONS e;
 	INDEXENTRY *ind;
@@ -421,6 +422,9 @@ int CoSave(UBYTE *inp)
 			c = *p; *p = 0;
 			if ( GetVar(inp,&type,&number,CEXPRESSION,NOAUTO) != NAMENOTFOUND ) {
 				if ( e[number].status == STOREDEXPRESSION ) {
+					if ( StrLen(AC.exprnames->namebuffer+e[number].name) > MAXENAME ) {
+						MesPrint("&Warning: saved expression name over %d char: %s", MAXENAME, AC.exprnames->namebuffer+e[number].name);
+					}
 /*
 		Here we have to locate the stored expression, copy its index entry
 		possibly after making a new fileindex and then copy the whole
@@ -518,11 +522,15 @@ NextExpr:;
 			,(LONG)sizeof(struct FiLeInDeX)) != (LONG)sizeof(struct FiLeInDeX) ) goto SavWrt;
 	}
 	else if ( !AP.preError ) {				/* All stored expressions should be saved. Easy */
+		/* Make sure there is at least one stored expression: */
 		if ( n > 0 ) { do {
-			if ( e->status == STOREDEXPRESSION ) break;
+			if ( StrLen(AC.exprnames->namebuffer+e->name) > MAXENAME ) {
+				MesPrint("&Warning: saved expr name over %d char: %s", MAXENAME, AC.exprnames->namebuffer+e->name);
+			}
+			if ( e->status == STOREDEXPRESSION ) exprInStorageFlag = 1;
 			e++;
 		} while ( --n > 0 ); }
-		if ( n ) {
+		if ( exprInStorageFlag ) {
 			wSize = TOLONG(AT.WorkTop) - TOLONG(AT.WorkPointer);
 			PUTZERO(scrpos);
 			SeekFile(AR.StoreData.Handle,&scrpos,SEEK_SET);		/* Start at the beginning */


jodavies:
Regarding Vlad's request for a mode which enforces the 16-char limit, I wonder: how about an `On strict;` mode which does this, but also can be used in the future for parser-related changes which make FORM more robust. In this mode, it would not be guaranteed that you have backward compatibility. If your old FORM scripts were relying on, essentially, buggy/undefined behaviour, they might break if you're using `On strict;`?

sources/store.c
```diff
@@ -421,6 +422,9 @@ int CoSave(UBYTE *inp)
 			c = *p; *p = 0;
 			if ( GetVar(inp,&type,&number,CEXPRESSION,NOAUTO) != NAMENOTFOUND ) {
 				if ( e[number].status == STOREDEXPRESSION ) {
+					if (StrLen(AC.exprnames->namebuffer+e[number].name) > MAXENAME) {
```
tueda, line 425: Maybe you can apply the same coding style as other if-statements (putting a space in parentheses inside conditional statements). The same for the other line with a similar if-statement.

sources/store.c
```diff
@@ -366,6 +366,7 @@ int CoSave(UBYTE *inp)
 	UBYTE *p, c;
 	WORD n = 0, i;
 	WORD error = 0, type, number;
+	WORD exprInStorage = 0;
```
tueda, line 369: `exprInStorage` is used as its type is `BOOL` (whether any expressions are in the storage or not). I wonder using `WORD` for this variable might look like it represents "the number of expressions in the storage" at first glance. The variable name could be changed to one that clearly suggests a boolean value or the type could be changed to `BOOL` (though the `BOOL` is not used in traditional (old) code).

tueda (commented):
Indeed I think this is a good addition. Later we can consider/refactor Warning/HighWarning and `On strict` in separate issues/PRs.

sources/store.c
jodavies: Right, this is not the first time I forgot this one...

sources/store.c
jodavies: Now I call it `exprInStorageFlag`, which should be more clear.

sources/store.c
```diff
@@ -524,13 +524,13 @@ NextExpr:;
 	else if ( !AP.preError ) {				/* All stored expressions should be saved. Easy */
 		/* Make sure there is at least one stored expression: */
 		if ( n > 0 ) { do {
-			if (StrLen(AC.exprnames->namebuffer+e->name) > MAXENAME) {
-				MesPrint("&Warning: saved expression name over %d char: %s", MAXENAME, AC.exprnames->namebuffer+e->name);
+			if ( StrLen(AC.exprnames->namebuffer+e->name ) > MAXENAME ) {
```
tueda, line 527: Sorry for nitpicking, but an extra space inside `StrLen(...)`, after `name`.

tueda:
I think this PR is good to merge, but before the merging maybe you would like to squash the commits into one (or I can do that). You may also write some test cases, which could be a separate PR.

jodavies:
A test case for this will be cleanest with #501 I think.

tueda:
OK, then I will merge this PR and also #501 (I don't see any problem with this).

jodavies:
This doesn't actually work quite as intended. If you use a loop to save multiple expressions, as in:
```
#do i = 1,2
   Save test-`i'.res, abcdefghijklmnop`i';
#enddo
```
you don't get two messages, but rather:
```
test-2.frm Line 18 --> Warning: saved expr name over 16 char: abcdefghijklmnop1
++++Errors in Loop
```

This seems to be due to the presence of the `&` at the beginning of the format string (which prints the filename and line number).

PR #502, opened 2024-04-30
Fix freeze in parsing complex conjugate operator

tueda:
The test case will come later (maybe after #501).


tueda:
In this version, the compiler, rather than the parser, generates the error. I also shortened the error message (which fits in one line, so the merged PR #504 was not mandatory for this PR at the end of the day) and added test cases.

@jodavies What do you think?

jodavies:
This looks fine to me, yes.

tueda:
@jodavies Thanks. I will merge it.

PR #503, opened 2024-04-30
Add test for PR #500

jodavies:
Also overlooked in PR: align the warning messages from the two cases.

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -1868,6 +1868,33 @@ for i in 1..3
   assert result("Hc#{i}") == result("Fc#{i}")
 end
 *--#] Issue191 : 
+*--#[ Issue192_1 :
+* Also related: Issue 334.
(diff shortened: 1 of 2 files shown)


jodavies:
By the way: I think it is OK to delete all of the branches from these merged PRs? There are quite a few in the repository now.

tueda (commented):
Thanks! I will merge it.

tueda:
To confirm, did you choose `expr` instead of `expression` just because of brevity? Or because the `warning?` method didn't work with long lines where newline characters are inserted? If it is the latter case, #504 provides the solution.

jodavies:
For brevity. I thought it better if it is more likely to fit on one line. Of course if you change the width or have sufficiently long expression names there will be linebreaks anyway.

tueda:
OK, then I will merge it (and then #504; there should be no more oversights in `warnings?` etc., I hope).

Issue #467, opened 2024-01-29
Testing FORM Project buildsystem to Crossmake


TaiXeflar:
Though I'm a not knowing about this project too(have seen the translated news about the developers), but I think it actually can have more users and have more tests on cross-platform. So I give it a try on @tueda's fork #266 and make some changes that completly using cmake to build Form 4.2. I tested about a midnight and have some tested news here.

The bad news is half and half: The compiled form executable can't run `check/example.frm` and other original `\*.frm` form language sources. Actually tested that whatever ways to build(Meaning building on Windows, on Linux, or using automake to build) are failed to run check examples, but example codes in Cookbook can run.

The good news is it can actually run the FORM compiler build task on Windows 11and Linux(I don't have macbook so ignore it). Tested that MSVC(`cl.exe`) and Clang are successful to build with Ninja buildsystem, makes total compilation only 3~4 seconds. 

Ready to share the result(CMakeLists.txt) and hope it helps!

Thanks
                TaiXeflar

PS: Screenshot on compile success and `\*.frm` form code test(If picture can show correctly)
![image](https://github.com/vermaseren/form/assets/81042947/0e9a4093-f619-4a71-82a4-efd3055190c7)

tueda:
Thank you for trying.

> The compiled form executable can't run `check/example.frm`

I am not sure about which command you invoked to run the example, but if you run
```
/path/to/build/sources/form check/example.frm
```
and get
```
...
~~~Use -D TEST=XXX
...
Program terminating at check/examples.frm Line 1 -->
...
```
then it means FORM terminated **as expected**; you need to choose a test case by `-D TEST=XXX` when running outside the testsuite. Note that some test cases cannot work without POSIX.

IMHO, portability is nice and CMake can be a good choice, possibly with a package manager like vcpkg to handle the dependencies. Ideally, the Autotools build might be completely replaced by a CMake build in future, but this is not realistic in one go. What we can do is to add an optional CMake build while keeping the Autotools build working. Although [your fork](https://github.com/TaiXeflar/form/tree/96099efb56165f1f09c8ad6a85d4d93cbe2532ef) may not be compatible with the existing Autotools build (and moving "optional" files may upset a major part of practical users who run the threaded version of FORM on Unix), your work is a good starting point to discuss it.

TaiXeflar:
[...]

PR #509, opened 2024-05-02
Fix memory leak in makeinteger

### LINKED ISSUE #508: MakeInteger: No memory while allocating NumberMalloc
Dear all,

Sadly in several instances I keep running into a bug when calling MakeInteger on a large number of terms, at some point getting back the message `No memory while allocating NumberMalloc`. From my idea of the inner workings of FORM MakeInteger should be basically cost-free in terms of memory, which makes me quite curious about this issue.

I suspect there is some memory leak in *MakeInteger in `src/argument.c`. It happens whenever I have factors of the function g and reach a certain number of terms with g that makeinteger acts upon.

With the standard FORM buffer sizes, I can replicate the error with the following simple FORM code. The number of terms here is set to the minimum for which the programme aborts using FORM 4.3.0:
```
cf f g;
s N;
on highfirst;
l F = 
#do i=1,524284
+ f(`i')*g(1+N)*g(N-1)
#enddo
;
makeinteger g;
.end
``` 
I would be very thankful for some insight.

Cheers,
Markus

jodavies:
We need to call NumberFree each loop iteration, not only upon leaving execarg.

This fixes #508 but I think it needs more testing.

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2668,3 +2668,26 @@ L F = #x;
 .end
 assert compile_error?("Illegal position for #")
 *--#] Issue499_2 :
+*--#[ Issue508 :
+Off statistics;
+cf f g;
+s N;
+on highfirst;
+l F =
+#do i=1,5000
++ f(`i')*(g(1+N/`i')*g(N/`i'-1))^25
+#enddo
+;
+.sort
+makeinteger g;
+id f(N?) = N^50;
+id g(N?) = 1;
+print;
+.end
+#require linux?
+#ulimit -v 8_000_000
+# We assume more memory than a 32bit system can provide
+#require wordsize >= 4
+assert succeeded?
+assert result("F") =~ expr("5000")
+*--#] Issue508 :
--- a/sources/argument.c
+++ b/sources/argument.c
@@ -66,7 +66,7 @@ WORD execarg(PHEAD WORD *term, WORD level)
 	WORD *oldwork = AT.WorkPointer, *oldwork2, scale, renorm;
 	WORD kLCM = 0, kGCD = 0, kGCD2, kkLCM = 0, jLCM = 0, jGCD, sign = 1;
 	int ii, didpolyratfun;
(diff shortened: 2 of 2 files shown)


jodavies:
Adding a test to the CI for this one is a bit tricky, because the number of terms required for the crash is system dependent.

jodavies:
This most recent commit seems like the best way to do this, but I don't see a performance difference compared to the commit before. I think NumberMalloc calls are just really cheap, and as long as you are not leaking and extending the buffer all of the time, it doesn't matter much how many you allocate.

check/fixes.frm
```diff
@@ -2655,3 +2655,23 @@ assert stdout =~ exact_pattern(<<'EOF')
...
++ f(`i')*(g(1+N/`i')*g(N/`i'-1))^25
+#enddo
+;
+.sort
+makeinteger g;
+id f(N?) = N^50;
+id g(N?) = 1;
+print;
+.end
+#ulimit -v 8_000_000
```
tueda, line 2674: Though tests on Windows in CI are currently run with bash, I feel using `#require unix?` (or `linux?`) with `#ulimit` is reasonable; it might not be so natural to assume bash will always be available on Windows, especially if alternative build methods ([CMake](https://github.com/vermaseren/form/issues/467) etc.) are implemented in future.

I think you also need `#require wordsize == 4` because the amount of memory usage on 64-bit machines is assumed.

tueda (commented):
To run the workflows, it seems necessary to rebase the PR in alignment with the current master branch.

tueda:
You can rebase this PR like this: [`tueda/jodavies_issue-508_rebased`](https://github.com/tueda/form/tree/jodavies_issue-508_rebased) (then `git diff` with 81862dc2567f2fd52b6c5da5cac59ab1e7f0b4c8 shows nothing).

tueda:
Right, in the end, squashing all to one or two commits is good. (I thought first rebasing is needed, but it seems that you can squash in one go.)

tueda:
Now this seems fine. I will revisit it tomorrow morning and probably will merge it.

PR #510, opened 2024-05-03
test(check.rb): add #ulimit instruction

tueda:
This may resolve the problem in https://github.com/vermaseren/form/pull/509#issuecomment-2090771016; "no memory" failures are system-dependent.

For example, in the following test fold:
```form
*--#[ example :
.end
#require linux?
#require wordsize == 4
#pend_if !serial?
#ulimit -v 4_000_000
assert succeeded?
*--#] example :
```
the maximum virtual memory is limited to `4*1000*1000*1024` ~ 4 GB (at least, bash and fish seem to support `ulimit -v`). If you decrease the memory size to `2_000_000`, then the FORM job fails due to insufficient memory.

With a fixed maximum memory size (and the same word size), it must be easy to reproduce a memory shortage in a certain situation.

@jodavies Can you try this?

### DIFF
--- a/check/check.rb
+++ b/check/check.rb
@@ -347,7 +347,7 @@ module FormTest
         nfiles.times do |i|
           ENV["FORM"] = FormTest.cfg.form_cmd
           @filename = "#{i + 1}.frm"
-          execute("#{FormTest.cfg.form_cmd} #{@filename}")
+          execute("#{ulimits}#{FormTest.cfg.form_cmd} #{@filename}")
           if !finished?
             info.status = "TIMEOUT"
             assert(false, "timeout (= #{timeout} sec) in #{@filename} of #{info.desc}")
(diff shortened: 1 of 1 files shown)


jodavies:
Yes this looks good. With a 4G limit you can crash with the version of the test in https://github.com/vermaseren/form/issues/508#issuecomment-2092414213 with i running to 5000, in 0.8s. This configuration causes a ~20GB virt.

tueda:
OK. Thanks. Then, I will merge it.

In fact, `ulimit` is shell-dependent. [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ulimit.html) only specifies the `-f` option. Maybe
```form
#require ENV["SHELL"] =~ /bash|fish/  # or ulimit -v probably works with dash/zsh etc.?
```
or something like that could be needed.

Issue #461, opened 2023-11-01
gcc-13 runtime error


apelloni:
I was building (t)form on my machine which is currently using `gcc (GCC) 13.2.1 20230801` and found an error.
The compilation is completed without any issues (except some warnings see form_build_gcc13.log)

Most programs run just fine, but I get an error with tables. By running
```
Table sparse,C(1);
.end
```
I get
```
test.frm Line 1 --> Illegal option in table definition: 'sparse,C(1)'
```

I've tried to build `form-5.0.0`, `form-4.3.1`, `form-4.3.0` but the issue is consistent.

It all disappear when I compile with `gcc-12 (GCC) 12.3.0` by running `CC=gcc-12 CXX=g++-12 ./configure`
Then the program runs as expected.

I've also tried to enable debugging but then `vorm` runs without issues.
So I've tried to compile with a lower optimization level and I found that `-O2` and `-O3` are the ones producing the issue

I'm unsure if this is a problem in the `gcc` compiler or something being exposed only at this level.
It could also be that the version I'm using is not completely stable.
I also didn't have the opportunity to try to reproduce it on another machine

I'm attaching also the output from `make form` using both `gcc-12` and `gcc-13`:
[form_build_gcc13.log](https://github.com/vermaseren/form/files/13229215/form_build_gcc13.log)
[form_build_gcc12.log](https://github.com/vermaseren/form/files/13229216/form_build_gcc12.log)

tueda:
Thank you for your report! I have confirmed the same FORM runtime error with GCC 13.2.0.

I am not sure whether GCC 13 finds a pitfall in FORM code (legal optimization with undefined behaviour) or it is just a bug of GCC, but it seems that making `c` volatile somehow removes the issue:
```diff
diff --git a/sources/names.c b/sources/names.c
index fd01e89..9ae586f 100644
--- a/sources/names.c
+++ b/sources/names.c
@@ -1666,7 +1666,8 @@ static int nwarntab = 1;
 int DoTable(UBYTE *s, int par)
 {
        GETIDENTITY
-       UBYTE *name, *p, *inp, c;
+       UBYTE *name, *p, *inp;
+       volatile UBYTE c;
        int i, j, k, sparseflag = 0, rflag = 0, checkflag = 0;
        int error = 0, ret, oldcbufnum, oldEside;
        WORD funnum, type, *OldWork, *w, *ww, *t, *tt, *flags1, oldnumrhs,oldnumlhs;
```

[...]

Issue #418, opened 2022-11-10
Wrong newline characters (CR+CR+LF) on Windows


tueda:
I don't yet fully understand it,  but it seems that on Windows (without POSIX; `UNIX` is undefined and `WINDOWS` is defined, which means `WITHRETURN` is defined), FORM tries to put `CR` before `LF` for newline characters. For example:
https://github.com/vermaseren/form/blob/e15da79ddec5dcb9432614e96854c7780496537c/sources/sch.c#L94-L97

But, in text mode,  `LF` written to output is converted to `CR+LF`, and `CR+LF` from input is converted to `LF`:
- https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen

In fact, `stdout` is opened with text mode:
- https://learn.microsoft.com/en-us/cpp/c-runtime-library/text-and-binary-mode-file-i-o

while other files in FORM are always opened in binary mode:
```
$ rg Uopen
sources/form3.h
485:extern FILES *Uopen(char *,char *);
512:#define Uopen(x,y) fopen(x,y)

sources/tools.c
988:    if ( ( f = Uopen(name,"rb") ) == 0 ) return(-1);
1007:   if ( ( f = Uopen(name,"a+b") ) == 0 ) return(-1);
1028:   if ( ( f = Uopen(name,"r+b") ) == 0 ) return(-1);
1047:   if ( ( f = Uopen(name,"w+b") ) == 0 ) return(-1);
1064:   if ( ( f = Uopen(name,"w+b") ) == 0 ) return(-1);

sources/unixfile.c
64:     #[ Uopen :
67:FILES *Uopen(char *filename, char *mode)
69:     FILES *f = (FILES *)Malloc1(sizeof(FILES),"Uopen");
86:     M_free(f,"Uopen");
91:     #] Uopen :
```
Consequently, FORM outputs to `stdout` contain wrong newline characters `CR+CR+LF` on Windows. This breaks some test cases where newline characters do matter.

tueda:
[Here](https://github.com/tueda/form/actions/runs/8814925434), I built Windows binaries with MSVC as well as MSYS2. Both give the same output. In the following, new-line characters are explicitly shown as `\r`  -> `<CR>` and `\n`  -> `<LF>\n`.

**Input:**
```form
S x,y;<CR><LF>
L F = (x+y)^10;<CR><LF>
P;<CR><LF>
.end
```

**FORM output to log file (`-l`)**
```
[...]

PR #511, opened 2024-05-04
Improve Windows build

### LINKED ISSUE #602: Simultaneous sort file creation by all worker threads
#527 added code to check if the sort file can be created in `ReserveTempFiles()`. In TFORM, all worker threads run this check simultaneously, resulting in [random failures on Windows](https://github.com/tueda/form/actions/runs/12427907412/job/34700352890#annotation:8:281) (I guess the file is created with an exclusive lock).

I think we don't need worker threads to check the sort file availability. Only the master thread should handle this.

### LINKED ISSUE #605: TFORM crashes in CleanUpSort when TempSortDir is unavailable
If the sort file can't be created at startup, `Terminate(-1)` is called (#527):

https://github.com/vermaseren/form/blob/8abec49f876deae1d13ad159519b301f0b140dc5/sources/startup.c#L807-L810

This eventually leads to a call to `CleanUpSort(0)`. But, at startup, `B` (inside `GETIDENTITY`) is `NULL`  in

https://github.com/vermaseren/form/blob/8abec49f876deae1d13ad159519b301f0b140dc5/sources/sort.c#L4803-L4808

So, `AN` causes a null pointer dereference.

This results in a segfault before printing an error message on Windows, giving a test failure.

tueda:
This PR consists of 3 patches to improve the building binaries on Windows:
- FORM source
- test cases
- others, including the testing framework and CI workflow

Though to some extent this is ready for a merge, there are still some remaining issues:

- `LF` without preceding `CR` remains in output (https://github.com/vermaseren/form/issues/418#issuecomment-2094032590).

- 2 failed tests.
  - `Issue135_3` (features.frm:879)
  - `Sta_Fill_1` (examples.frm:647)

Concerning the 2 test failures, both are caused by segfaults (`Sta_Fill_1` leads to a segfault also on Linux). The unexpected failures might be related to the fact that MSYS2 GCC is currently version 13.2.0, the [buggy version](https://github.com/vermaseren/form/issues/461) (MSYS2 has [no way](https://github.com/msys2/setup-msys2/issues/287) to install a specific version like gcc 12).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -89,7 +89,7 @@ jobs:
             mingw-w64-x86_64-gmp
             mingw-w64-x86_64-mpfr
             mingw-w64-x86_64-zlib
-            ruby
+            mingw-w64-x86_64-ruby
 
       - name: Download tarball
         uses: actions/download-artifact@v4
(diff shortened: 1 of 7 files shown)


tueda:
Rebased onto master.

The failure of `Issue135_3` seems to have resolved itself, likely because of an update of the gcc version in the upstream (MSYS2).

The failure of `Sta_Fill_1` may not be the fault of the compiler or Windows. It seems to contain intrinsically undefined behaviour. Actually, Valgrind (on Linux) shows stack overflows:
```
FORM 5.0.0-beta.1 (May 27 2024, v5.0.0-beta.1-69-gc571f77)  Run: Mon May 27 14:48:08 2024
        Table B(1:1);
        Local dummy = 1;
        .sort

Time =       0.04 sec    Generated terms =          1
           dummy         Terms in output =          1
                         Bytes used      =         20
        Fill B(1) = dummy;
1.frm Line 4 --> Warning: Use of expressions and/or $variables in Fill statemen
ts is potentially very dangerous.
        Drop dummy;
        .sort
        Local F = B(1);
        Print;
        .end
==73714== Stack overflow in thread #1: can't grow stack to 0x1ffe801000
==73714== Stack overflow in thread #1: can't grow stack to 0x1ffe801000
==73714== Can't extend stack to 0x1ffe801078 during signal delivery for thread 1:
==73714==   no stack segment
==73714==
==73714== Process terminating with default action of signal 11 (SIGSEGV)
==73714==  Access not within mapped region at address 0x1FFE801078
==73714== Stack overflow in thread #1: can't grow stack to 0x1ffe801000
==73714==    at 0x1DF3F1: Generator (proces.c:3246)
==73714==  If you believe this happened as a result of a stack
==73714==  overflow in your program's main thread (unlikely but
==73714==  possible), you can try to increase the size of the
==73714==  main thread stack using the --main-stacksize= flag.
==73714==  The main thread stack size used in this run was 8388608.
```

tueda:
FYI, here is an experimental branch working with MSVC: https://github.com/tueda/form/tree/ci-msvc

jodavies:
`Sta_Fill_1` results in an infinite loop of `DoOnePow`, `Generator`, `Generator`, `Generator`. I suppose the table entry references the expression "dummy", but in `DoOnePow` we are using `nexp=0` which at this point refers to `F`. Hence the loop? FORM at least prints the warning.

So this one is nothing to do with windows, anyway.

tueda:
Rebased.

tueda:
Rebased.

tueda:
Rebased.

tueda:
Now there are no test failures on native Windows (except for the occasional random failures in TFORM).

Issue #468, opened 2024-02-21
Crash in MergePatches


jodavies:
Hello,

The attached script crashes in MergePatches (at sort.c L3604, with par=2, v4.3.1 tag) most of the time with tform.
Increasing the sub-buffer sizes fixes this example. The input expression is "small-ish"...
[crash.tar.gz](https://github.com/vermaseren/form/files/14361036/crash.tar.gz)

In some cases, eg `tvorm -w2` I can actually get this to not crash, but produce nonsense output (i.e, lots of bare terms, not inside the polyratfun.)

Thanks,
Josh.

jodavies:
The problem here arises in the sorting of the polyratfun arguments, like so:
```
Thread 5 "tvorm-test" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff33fff640 (LWP 550611)]
MergePatches (par=par@entry=2) at sort.c:3655
3655            *rr = 0;
(gdb) bt
#0  MergePatches (par=par@entry=2) at sort.c:3655
#1  0x0000555555685196 in EndSort (B=0x7ffc4c000b70, buffer=0x7ffed8323024, par=1) at sort.c:907
#2  0x0000555555635de1 in poly_sort (B=B@entry=0x7ffc4c000b70, a=a@entry=0x7ffed832301c) at polywrap.cc:573
#3  0x000055555563b276 in poly_ratfun_add (B=B@entry=0x7ffc4c000b70, t1=t1@entry=0x7fff3fda3bec, t2=t2@entry=0x7fff6da0fde0) at polywrap.cc:681
#4  0x00005555556b947e in SortBotMerge (B=B@entry=0x7ffc4c000b70) at threads.c:4322
#5  0x00005555556ba699 in RunSortBot (dummy=<optimised out>) at threads.c:1964
#6  0x00007ffff778fac3 in start_thread (arg=<optimised out>) at ./nptl/pthread_create.c:442
#7  0x00007ffff7821850 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```
The crash is due to `AR.CompressPointer` being null. It reminds me of #211 , but those problems were due to stage-sorting in the sub buffers, but in this example I think we did not reach that stage?

So, in this sort where should `AR.CompressPointer` have been set?

jodavies:
This example has two failure modes, actually. The one in the previous comment is fixed by this commit, I think:
[...]

PR #513, opened 2024-05-04
Better errors when (Sub)SmallExtension is filled.

### LINKED ISSUE #512: Bad error message for full SubSmallExtension
If you fill the `SubSmallExtension`:
```
* Sort which fills (sub)smallextension:

#: SmallSize 1000K
#: SmallExtension 1200K
#: TermsInSmall 2M
#: SubSmallSize 1000K
#: SubSmallExtension 1200K
#: SubTermsInSmall 2M

Symbol x,n;
CFunction f,g,prf;
CFunction hold;

Local test = 1;
.sort

PolyRatFun prf;
Term;
   Multiply (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(100)>);
   Identify f(x?) = prf(n-x,n+x);
EndTerm;

.end
```
FORM crashes with `Please increase SmallExtension in form.set`. Of course, increasing `SmallExtension` will never get you anywhere in this case.

I've not yet worked out how to determine here
https://github.com/vermaseren/form/blob/b82093350ae986ad6c65a9ad25bf748efa53d7be/sources/sort.c#L2236
whether we are doing a regular or sub-buffer sort. If this information is known, the error message can be printed with "Sub", accordingly.

Also the user may not be using a `form.set` file, maybe the error should just refer generically to the "(Sub)SmallExtension setup parameter".

jodavies:
Previously, FORM warns only about SmallExtension, even if it is a sub-buffer sort.

This prompts the user to adjust the wrong buffer in the setup.

Fixes #512

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2725,9 +2725,10 @@ EOF
 #: SortIOSize 200K
 #: SubSortIOSize 200K
 
-#: SubSmallSize 100K
-#: SubSmallExtension 200K
-#: SubTermsInSmall 5K
+#: SubLargeSize 134400480
+#: SubSmallSize 12800016
+#: SubSmallExtension 19200032
+#: SubTermsInSmall 5008
 
 #define N "30"
 
@@ -2845,19 +2846,17 @@ assert result("F") =~ expr("5000")
 #-
 * Sort which fills SmallExtension:
 
-#: SmallSize 1000K
-#: SmallExtension 1200K
-#: SubSmallSize 1000K
-#: SubSmallExtension 1200K
-#: SubTermsInSmall 2M
+* These are the smallest buffer sizes that are OK for -w4 tform workers
+#: SmallSize 10240064
+#: SmallExtension 15360096
+#: TermsInSmall 100K
 
 Symbol x,n;
-CFunction f,g,prf;
+CFunction g,f,prf;
 
-Local test = (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(100)>);
+Local test = (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(350)>);
 .sort
 
-Off parallel;
 PolyRatFun prf;
 
 Identify f(x?) = prf(n-x,n+x);
@@ -2873,13 +2872,11 @@ assert runtime_error?("Please increase SmallExtension setup parameter.")
 #-
 
 * Sort which fills SubSmallExtension:
-
-#: SmallSize 1000K
-#: SmallExtension 1200K
-#: TermsInSmall 2M
-#: SubSmallSize 1000K
-#: SubSmallExtension 1200K
-#: SubTermsInSmall 2M
+* These are the default sizes at the time of writing:
+#: SubSmallSize 2560016
+#: SubSmallExtension 3840032
+* These are not default:
+#: SubTermsInSmall 100K
 
 Symbol x,n;
 CFunction f,g,prf;
@@ -2904,18 +2901,14 @@ assert runtime_error?("Please increase SubSmallExtension setup parameter.")
 #-
 
 * Sort which fits in SmallExtension, but needs GarbHand
-
-#: SmallSize 1000K
-#: SmallExtension 5090K
-#: TermsInSmall 2M
-#: SubSmallSize 1000K
-#: SubSmallExtension 1200K
-#: SubTermsInSmall 2M
+#: SmallSize 10240064
+#: SmallExtension 20360K
+#: TermsInSmall 100K
 
 Symbol x,n;
-CFunction f,g,prf;
+CFunction g,f,prf;
 
-Local test = (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(100)>);
+Local test = (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(350)>);
 .sort
 
 PolyRatFun prf;
@@ -2930,13 +2923,11 @@ assert succeeded?
 #-
 
 * Sort which fits in SubSmallExtension, but needs GarbHand
-
-#: SmallSize 1000K
-#: SmallExtension 1200K
-#: TermsInSmall 2M
-#: SubSmallSize 1000K
+* These are the default sizes at the time of writing:
+#: SubSmallSize 2560016
+* These are not default:
 #: SubSmallExtension 5090K
-#: SubTermsInSmall 2M
+#: SubTermsInSmall 100K
 
 Symbol x,n;
 CFunction f,g,prf;
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2841,6 +2841,65 @@ print;
 assert succeeded?
 assert result("F") =~ expr("5000")
 *--#] Issue508 :
+*--#[ Issue512_1 :
+#-
+* Sort which fills SmallExtension:
+
+#: SmallSize 1000K
+#: SmallExtension 1200K
+#: SubSmallSize 1000K
+#: SubSmallExtension 1200K
+#: SubTermsInSmall 2M
+
+Symbol x,n;
+CFunction f,g,prf;
+
+Local test = (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(100)>);
+.sort
+
+Off parallel;
+PolyRatFun prf;
+
+Identify f(x?) = prf(n-x,n+x);
+
+.end
+# Fails due to polynomial size on 32bit builds
+#require wordsize >= 4
+# Runtime errors may freeze ParFORM.
+#pend_if mpi?
+assert runtime_error?("Please increase SmallExtension setup parameter.")
+*--#] Issue512_1 :
+*--#[ Issue512_2 :
+#-
+
+* Sort which fills SubSmallExtension:
+
+#: SmallSize 1000K
+#: SmallExtension 1200K
+#: TermsInSmall 2M
+#: SubSmallSize 1000K
+#: SubSmallExtension 1200K
+#: SubTermsInSmall 2M
+
+Symbol x,n;
+CFunction f,g,prf;
+
+Local test = 1;
+.sort
+
+PolyRatFun prf;
+Term;
+	Multiply (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(100)>);
+	Identify f(x?) = prf(n-x,n+x);
+EndTerm;
+
+.end
+# Fails due to polynomial size on 32bit builds
+#require wordsize >= 4
+# Runtime errors may freeze ParFORM.
+#pend_if mpi?
+assert runtime_error?("Please increase SubSmallExtension setup parameter.")
+*--#] Issue512_2 :
 *--#[ Issue525 :
 #:threadbucketsize 5
 #:processbucketsize 5
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2900,6 +2900,61 @@ EndTerm;
 #pend_if mpi?
 assert runtime_error?("Please increase SubSmallExtension setup parameter.")
 *--#] Issue512_2 :
+*--#[ Issue512_3 :
+#-
+
+* Sort which fits in SmallExtension, but needs GarbHand
+
+#: SmallSize 1000K
+#: SmallExtension 5090K
+#: TermsInSmall 2M
+#: SubSmallSize 1000K
+#: SubSmallExtension 1200K
+#: SubTermsInSmall 2M
+
+Symbol x,n;
+CFunction f,g,prf;
+
+Local test = (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(100)>);
+.sort
+
+PolyRatFun prf;
+Identify f(x?) = prf(n-x,n+x);
+
+.end
+# Fails due to polynomial size on 32bit builds
+#require wordsize >= 4
+assert succeeded?
+*--#] Issue512_3 :
+*--#[ Issue512_4 :
+#-
+
+* Sort which fits in SubSmallExtension, but needs GarbHand
+
+#: SmallSize 1000K
+#: SmallExtension 1200K
+#: TermsInSmall 2M
+#: SubSmallSize 1000K
+#: SubSmallExtension 5090K
+#: SubTermsInSmall 2M
+
+Symbol x,n;
+CFunction f,g,prf;
+
+Local test = 1;
+.sort
+
+PolyRatFun prf;
+Term;
+	Multiply (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(100)>);
+	Identify f(x?) = prf(n-x,n+x);
+EndTerm;
+
+.end
+# Fails due to polynomial size on 32bit builds
+#require wordsize >= 4
+assert succeeded?
+*--#] Issue512_4 :
 *--#[ Issue525 :
 #:threadbucketsize 5
 #:processbucketsize 5
--- a/sources/sort.c
+++ b/sources/sort.c
@@ -2234,11 +2234,21 @@ WORD AddPoly(PHEAD WORD **ps1, WORD **ps2)
 					TalToLine((UWORD)(*s2++)); TokenToLine((UBYTE *)"  ");
 				}
 				FiniLine();
-				MesPrint("Please increase SmallExtension in %s",setupfilename);
+				if ( AR.sLevel > 0 ) {
+					MesPrint("Please increase SubSmallExtension setup parameter.");
+				}
+				else {
+					MesPrint("Please increase SmallExtension setup parameter.");
+				}
 				MUNLOCK(ErrorMessageLock);
 #else
 				MLOCK(ErrorMessageLock);
-				MesPrint("Please increase SmallExtension in %s",setupfilename);
+				if ( AR.sLevel > 0 ) {
+					MesPrint("Please increase SubSmallExtension setup parameter.");
+				}
+				else {
+					MesPrint("Please increase SmallExtension setup parameter.");
+				}
 				MUNLOCK(ErrorMessageLock);
 #endif
 				Terminate(-1);
--- a/sources/sort.c
+++ b/sources/sort.c
@@ -3315,7 +3315,7 @@ LONG SplitMerge(PHEAD WORD **Pointer, LONG number)
 {
 	GETBIDENTITY
 	SORTING *S = AT.SS;
-	WORD **pp3, **pp1, **pp2;
+	WORD **pp3, **pp1, **pp2, **pptop;
 	LONG i, newleft, newright, split;
 
 	if ( number < 2 ) return(number);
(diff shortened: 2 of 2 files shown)


jodavies:
It looks like `parform` hangs after printing the error here.

check/fixes.frm
```diff
@@ -2668,3 +2668,54 @@ L F = #x;
...
+Local test = (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(100)>);
+.sort
+
+Off parallel;
+PolyRatFun prf;
+
+Identify f(x?) = prf(n-x,n+x);
+
+.end
+assert runtime_error?("Please increase SmallExtension setup parameter.")
```
tueda, line 2693: Can you add the following code snippet for ParFORM (after confirming that ParFORM displayed the correct runtime errors in GitHub Actions)?
```
# Runtime errors may freeze ParFORM.
#pend_if mpi?
```

check/fixes.frm
```diff
@@ -2668,3 +2668,54 @@ L F = #x;
...
+.sort
+
+PolyRatFun prf;
+Term;
+	Multiply (<f(1)>+...+<f(150)>)*(<g(1)>+...+<g(100)>);
+	Identify f(x?) = prf(n-x,n+x);
+EndTerm;
+
+.end
+assert runtime_error?("Please increase SubSmallExtension setup parameter.")
```
tueda, line 2720: Because `Issue512_2` failed on 32-bit containers (`check-i386 (form/tform)`), you also need to add the following code to prevent it from running on 32-bit machines:
```
#require wordsize >= 4
```

tueda:
Hmm...For some reason `tform` on `macos-11` doesn't print `"Please increase SubSmallExtension setup parameter."` before terminating.

jodavies:
I have no way to test if it crashes for some other reason.

tueda:
This may be irrelevant, but on MacOS (`Sonoma 14.4.1`, `Apple M2`, `Apple clang version 15.0.0`), I got some `malloc` errors for `Issue512_2`:
```bash
./build/sources/vorm -D TEST=Issue512_2 check/fixes.frm 
```
```
FORM 5.0.0-beta.1 (May  4 2024, v5.0.0-beta.1-57-g1a75289)  Run: Sat May 11 17:47:37 2024 
    #ifndef `TEST' 
      #message Use -D TEST=XXX 
      #terminate 
    #else 
      #include `NAME_' # `TEST' 
    #- 
 
Time =       0.00 sec    Generated terms =          1 
            test         Terms in output =          1 
                         Bytes used      =         20 
Please increase SubSmallExtension setup parameter. 
Program terminating at check/fixes.frm Line 2722 -->  
  0.12 sec out of 0.13 sec 
vorm(25084,0x1fcafbac0) malloc: *** error for object 0x300000001: pointer being freed was not allocated 
vorm(25084,0x1fcafbac0) malloc: *** set a breakpoint in malloc_error_break to debug 
```
The corresponding Valgrind run on Linux is:
```bash
valgrind ./build/sources/vorm -D TEST=Issue512_2 check/fixes.frm
```
```
FORM 5.0.0-beta.1 (May  4 2024, v5.0.0-beta.1-57-g1a75289)  Run: Sat May 11 17:55:02 2024
...
Time =       0.05 sec    Generated terms =          1
            test         Terms in output =          1
                         Bytes used      =         20
==118620== Invalid write of size 4
==118620==    at 0x20C263: GarbHand (sort.c:3599)
==118620==    by 0x20CA39: AddPoly (sort.c:2199)
==118620==    by 0x20CD2F: SplitMerge (sort.c:3345)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20CBBB: SplitMerge (sort.c:3331)
==118620==    by 0x20CBBB: SplitMerge (sort.c:3331)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20E7EB: StoreTerm (sort.c:4440)
==118620==    by 0x1DF55F: Generator (proces.c:3366)
==118620==  Address 0x4bf8a5b8 is 0 bytes after a block of size 22,857,080 alloc'd
==118620==    at 0x4849839: malloc (in /home/linuxbrew/.linuxbrew/Cellar/valgrind/3.22.0/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==118620==    by 0x22C03F: Malloc1 (tools.c:2322)
==118620==    by 0x206335: AllocSort (setfile.c:916)
==118620==    by 0x209261: NewSort (sort.c:641)
==118620==    by 0x126FDB: CoIdExpression (comexpr.c:824)
==118620==    by 0x12765A: CoId (comexpr.c:398)
==118620==    by 0x13ADF2: CompileStatement (compiler.c:658)
==118620==    by 0x1D76E1: PreProcessor (pre.c:1123)
==118620==    by 0x214BAE: main (startup.c:1690)
==118620==
Please increase SubSmallExtension setup parameter.
Program terminating at check/fixes.frm Line 2722 -->
==118620== Invalid read of size 4
==118620==    at 0x21443B: Terminate (startup.c:1792)
==118620==    by 0x4BF78A6F: ???
==118620==    by 0x20CA86: AddPoly (sort.c:2249)
==118620==    by 0x20CD2F: SplitMerge (sort.c:3345)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20CBBB: SplitMerge (sort.c:3331)
==118620==    by 0x20CBBB: SplitMerge (sort.c:3331)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==    by 0x20E7EB: StoreTerm (sort.c:4440)
==118620==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
==118620==
  6.78 sec out of 6.79 sec
==118620== Invalid free() / delete / delete[] / realloc()
==118620==    at 0x484C89F: free (in /home/linuxbrew/.linuxbrew/Cellar/valgrind/3.22.0/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==118620==    by 0x22C0B0: M_free (tools.c:2449)
==118620==    by 0x1404D8: ClearSortGZIP (compress.c:731)
==118620==    by 0x20FCDC: CleanUpSort (sort.c:4737)
==118620==    by 0x2140EB: CleanUp (startup.c:1710)
==118620==    by 0x21440E: Terminate (startup.c:1850)
==118620==    by 0x214A53: onErrSig (startup.c:1558)
==118620==    by 0x4D8E51F: ??? (in /usr/lib/x86_64-linux-gnu/libc.so.6)
==118620==    by 0x21443A: Terminate (startup.c:1790)
==118620==    by 0x4BF78A6F: ???
==118620==    by 0x20CA86: AddPoly (sort.c:2249)
==118620==    by 0x20CD2F: SplitMerge (sort.c:3345)
==118620==  Address 0x1f00000098 is not stack'd, malloc'd or (recently) free'd
==118620==
==118620== Invalid free() / delete / delete[] / realloc()
==118620==    at 0x484C89F: free (in /home/linuxbrew/.linuxbrew/Cellar/valgrind/3.22.0/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==118620==    by 0x22C0B0: M_free (tools.c:2449)
==118620==    by 0x1404E8: ClearSortGZIP (compress.c:732)
==118620==    by 0x20FCDC: CleanUpSort (sort.c:4737)
==118620==    by 0x2140EB: CleanUp (startup.c:1710)
==118620==    by 0x21440E: Terminate (startup.c:1850)
==118620==    by 0x214A53: onErrSig (startup.c:1558)
==118620==    by 0x4D8E51F: ??? (in /usr/lib/x86_64-linux-gnu/libc.so.6)
==118620==    by 0x21443A: Terminate (startup.c:1790)
==118620==    by 0x4BF78A6F: ???
==118620==    by 0x20CA86: AddPoly (sort.c:2249)
==118620==    by 0x20CD2F: SplitMerge (sort.c:3345)
==118620==  Address 0x41fffffff0 is not stack'd, malloc'd or (recently) free'd
==118620==
==118620== Syscall param unlink(pathname) points to unaddressable byte(s)
==118620==    at 0x4E620AB: unlink (syscall-template.S:120)
==118620==    by 0x4DADE6C: remove (remove.c:33)
==118620==    by 0x20FCF4: CleanUpSort (sort.c:4741)
==118620==    by 0x2140EB: CleanUp (startup.c:1710)
==118620==    by 0x21440E: Terminate (startup.c:1850)
==118620==    by 0x214A53: onErrSig (startup.c:1558)
==118620==    by 0x4D8E51F: ??? (in /usr/lib/x86_64-linux-gnu/libc.so.6)
==118620==    by 0x21443A: Terminate (startup.c:1790)
==118620==    by 0x4BF78A6F: ???
==118620==    by 0x20CA86: AddPoly (sort.c:2249)
==118620==    by 0x20CD2F: SplitMerge (sort.c:3345)
==118620==    by 0x20CBA2: SplitMerge (sort.c:3330)
==118620==  Address 0x5 is not stack'd, malloc'd or (recently) free'd
```
Something wrong seems to be happening before calling `Terminate()`...

jodavies:
I'll have a look at it. It might be related to https://github.com/vermaseren/form/issues/468#issuecomment-2104900922

jodavies:
The valgrind error here is due to `GarbHand` making no attempt to avoid writing out of bounds. The old content of the small buffer, plus the stuff in SplitScratch, is copied back to the small buffer here:
https://github.com/vermaseren/form/blob/0112aa6b21b48616575c6cd205ec00c7797dbc93/sources/sort.c#L3581-L3611

The function certainly knows things have gone wrong; if you define `TESTGARB` it prints "We are in deep trouble", and form is certainly going to `Terminate` after return. But this can be made "valgrind clean" by comparing `s2+i` with `sTop2` in each iteration of the while loops (maybe just the second: the original content of the small buffer should certainly fit again), before the `NCOPY`.

That we get a valgrind error here means that the content of the old small buffer + SplitScratch has blown through the small extension, and also through the `PObuffer`, which follows the small buffer in the allocation of `AllocSort`.

jodavies:
>(maybe just the second: the original content of the small buffer should certainly fit again)

This is actually not the case here; it is the first loop which causes the problem. I don't understand why, though. The general idea is,
1. In `AddCoef` and `AddPoly`, try to put a new term into the small buffer
2. If it doesn't fit within the SmallExtension (sTop2), call GarbHand
3. Check total size of existing sPointer and SplitScratch terms. If they fit in the spare space in the LargeBuffer use that, otherwise allocate space
4. Copy the terms of sPointer and SplitScratch to the temporary space, skipping holes.
5. Copy terms back to the SmallBuffer+Extension, (again skipping holes?)

I would expect that the original terms pointed to by sPointer should be guaranteed to fit in the SmallBuffer+Extension?

jodavies:
I added some prints to try to work out what is going on here. For a "working" call of `GarbHand`, I get something like the following:
```
sort.c AddPoly 2196, make new term (S->sFill 932624) (new term 7106) --> 939730
sort.c AddPoly 2196, make new term (S->sFill 939730) (new term 9192) --> 948922
sort.c AddPoly 2196, make new term (S->sFill 948922) (new term 12614) --> 961536
------Garbage collection-------
in:  S->sFill = 948922, S->sTop2 = 960008
total = 926812, nterms = 0, SmallSize = 640004, SmallESize = 960008
out:  S->sFill = 926812, S->sTop2 = 960008
```
so as soon as sFill plus the size of the new term is >= sTop2, GarbHand is called which clears out the gaps in the small buffer (this is all during a SplitMerge so terms are being merged a lot).

At the problematic call:
```
sort.c AddPoly 2196, make new term (S->sFill 959886) (new term 72) --> 959958
sort.c AddPoly 2196, make new term (S->sFill 959958) (new term 48) --> 960006
sort.c AddPoly 2196, make new term (S->sFill 960006) (new term 48) --> 960054
------Garbage collection-------
in:  S->sFill = 960006, S->sTop2 = 960008
total = 982988, nterms = 0, SmallSize = 640004, SmallESize = 960008
==96593== Invalid write of size 4
==96593==    at 0x221EB0: GarbHand (sort.c:3597)
==96593==    by 0x22277E: AddPoly (sort.c:2206)
==96593==    by 0x222B07: SplitMerge (sort.c:3321)
```
Here "total" is larger than sTop2 (I changed the print compared to the master branch, which prints 2*total for some reason). But sFill was 960006 before the call, and it was only trying to add a term of 48 words. So I am not sure how a "total" of 982988 is possible, unless the `sPointer` array contains terms unaccounted for in `sFill` -- as I understand, this should not be the case.

I am testing on top of #529 so writes over sTop2 are an immediate crash, unlike in the master branch which needs a write beyond PObuffer also (the test rebased onto master actually doesn't crash anymore, due to the increased value of `SIOsize` from #515).

jodavies:
The issue is that "NEWSPLITMERGE" is not strict about setting pointers to zero, for terms which have merged (or cancel, but in this test nothing cancels). If we use the old code, `GarbHand` correctly computes the size of the terms which remain.

Maybe this was done for speed, but what is means is that the "new" code fails in some cases where the new succeeds because GarbHand cannot work properly. If you run `Issue512_2` with a `SubSmallExtension` of 5090K, it will run to completion with the old code, but crash with the new because it determines (incorrectly) that SubSmallExtension is too small.. The `Timsort` code is commented for this tests, but it doesn't change anything.

jodavies:
@tueda can you test this on your mac?

In case it is needed again, I used the following print code (and diff) in the two versions of `SplitMerge`, to make sure the pointer array now looks the same in both cases.
```
// Print current state
printf("%4ld: ", number);
for (int ii = 0; ii < S->sTerms; ii++) {
       if ( (S->sPointer)[ii] ) {
               printf("X");
       }
       else {
               printf(".");
       }
}
printf("\n");
fflush(stdout);
```

tueda:
Now it works on my MacBook Air.

jodavies:
OK, that is good. I have run a few mincer benchmarks and did not see any performance impact. I think in FORM the term comparison is so expensive that no amount of pointer operations during the sort is noticable.

I have a "re-written" version of `SplitMerge` with some variable name changes and a bunch of commentary to make sure I understand it. Maybe the "improvement in the style of Timsort" can be further improved; at the moment it checks the right-hand half of the left array for the merge point of the right array with a binary search. In principle one could check the entire left array for the merge point, instead of just giving up immediately if the terms are not ordered at the midpoint. Maybe Jos already tested this at the time, and found it not to help? I will test this, but it would be for a new PR.

jodavies:
I missed some pointer zeroing which was causing GarbHand to still double-count terms. The crash was due to the merge of the split sort allocation, so now errors here are overruns of the small extension buffer (this passed the CI before, where the overruns were just spilling into the PO buffer).

The "Timsort" part needed to be modified slightly, when there are not two copies of the terms present, but it is functionally the same.

I will do some performance regression checks later.

jodavies:
I don't see any performance impact due to this compared to master, and indeed I can see a ~1% improvement due to the Timsort code (I could not measure a difference there in the past). I'll merge the fix commit into the previous so that there isn't a buggy commit on the master branch.

tueda:
Before merging, can you remove warnings in the test cases added by this PR? You can see them by running:
```
./check/check.rb ./build/sources/vorm 'Issue512*' -v
```

These warning messages are not caught by the test suite for now (https://github.com/vermaseren/form/issues/581) but should be resolved.

jodavies:
Yes, this is just a case of using the buffer sizes that form uses anyway.

I could also add a "debug mode" for splitmerge which I used to test the changes (hidden behind a preproc variable) which could be useful in the future?

tueda:
You mean debug code disabled by a C preprocessor #ifdef directive? Yes, that would be useful.

tueda:
So, the problem is that the buffer size should be adjusted for the number of workers, right?

One possible (and tedious) solution is to make several test cases for, e.g., `form` & `parform`, `tform -w2` and `tform -3`, by setting `#require serial? || mpi?`, `#require threaded? && ncpu == 2` and `#require threaded? && ncpu == 4`.

A more sophisticated solution would be to introduce a kind of template
```
#: SmallSize #{some_Ruby_expression}K
```
but it requires some work in `check.rb`.

jodavies:
> So, the problem is that the buffer size should be adjusted for the number of workers, right?

Yes and no. With sizes per-worker, the single-threaded valgrind runs could be made smaller and faster, but the `-w4` test is going to be the bottleneck regardless, if the test is configured to test the intended thing for tform.

jodavies:
OK, then I will get the proper tests in and see how they run in the CI. It could run a bit faster in general, by splitting the tests into a larger number of groups -- usually the CI is waiting for just a few (<20, the limit) tests to finish for a while.

jodavies:
Now if you define `SPLITMERGEDEBUG` in `sort.c`, you will get an output which shows the pointers into the small buffer on entry to `SplitMerge`.
```
FORM 5.0.0-beta.1 (Nov 18 2024, v5.0.0-beta.1-85-g4411553-dirty)  Run: Mon Nov 18 15:25:27 2024
    Symbol x;

    Local test = (1+x)^10;
    .sort
  11:    0    4   12   20   28   36   44   52   60   68   76
   5:    0    4   12   20   28   36   44   52   60   68   76
   2:    0    4   12   20   28   36   44   52   60   68   76
   3:    0    4   12   20   28   36   44   52   60   68   76
   1:    0    4   12   20   28   36   44   52   60   68   76
   2:    0    4   12   20   28   36   44   52   60   68   76
   6:    0    4   12   20   28   36   44   52   60   68   76
   3:    0    4   12   20   28   36   44   52   60   68   76
   1:    0    4   12   20   28   36   44   52   60   68   76
   2:    0    4   12   20   28   36   44   52   60   68   76
   3:    0    4   12   20   28   36   44   52   60   68   76
   1:    0    4   12   20   28   36   44   52   60   68   76
   2:    0    4   12   20   28   36   44   52   60   68   76

Time =       0.00 sec    Generated terms =         11
            test         Terms in output =         11
                         Bytes used      =        268

    Identify x = -1;
   1:    0

    Print;
    .end
  11:    0    4    8   12   16   20   24   28   32   36   40
   5:    0    4    8   12   16   20   24   28   32   36   40
   2:    0    4    8   12   16   20   24   28   32   36   40
   3:    0 ....    8   12   16   20   24   28   32   36   40
   1:    0 ....    8   12   16   20   24   28   32   36   40
   2:    0 ....    8   12   16   20   24   28   32   36   40
   6:    0 .... .... .... ....   20   24   28   32   36   40
   3:    0 .... .... .... ....   20   24   28   32   36   40
   1:    0 .... .... .... ....   20   24   28   32   36   40
   2:    0 .... .... .... ....   20   24   28   32   36   40
   3:    0 .... .... .... ....   20 .... ....   32   36   40
   1:    0 .... .... .... ....   20 .... ....   32   36   40
   2:    0 .... .... .... ....   20 .... ....   32   36   40

Time =       0.00 sec    Generated terms =         11
            test         Terms in output =          0
                         Bytes used      =          4

   test = 0;

  0.00 sec out of 0.00 sec
```
This produces huge output for larger problems, but it allows you do diff the logs and make sure things are going OK. You should get the same output, regardless of your `NEWSPLITMERGE` or `NEWSPLITMERGETIMSORT` settings.

In the master branch if you put the same printing code, you will see for the Issue512 test cases (best not run with `tform`...) that the array contains some non-zero pointers in the NEWSPLITMERGE mode, compared to the original code. It is these non-zero pointers which were causing the issues in `GarbHand`.

jodavies:
@tueda if you are happy with this I would like to merge it, a collaborator has smallextension problems with some jobs and this should fix it for them.

tueda:
Yes, I think this can be merged. I'll merge it.

jodavies:
Thanks. This one should go into 4.3 as well, if you are not already on it.

tueda:
Yes, I'm trying to resolve conflicts.

Issue #292, opened 2018-08-23
(1)Output should fit inside a single term. Increase MaxTermSize?


jodavies:
Hello,
This relates to the discussion in https://github.com/vermaseren/form/pull/289#issuecomment-415432136
I have managed to extract a minimal crashing example from a larger set of codes: (rename as .frm)
[test-minimal.txt](https://github.com/vermaseren/form/files/2315449/test-minimal.txt)

The example has two modes, inside a function argument (`#define DOINARG') or not.

Outside the function argument, everything runs OK. It also checks in each loop whether the current expression does indeed fit inside a function argument; it does.

Computing inside a function argument, however, it always crashes, however large I set MaxTermSize (10M, for eg, does not work).

For the first (commented) example input, it works for MaxTermSize values larger than 42K. With the default, (with my patches), I get the output
```
Sorted function argument too long: 41004 words
MaxTermSize: 40000 words
Called from execarg
```
which is a different error from the crashing second input term.

The code which this has been taken from has run successfully on many millions of very similar input terms. This one causes problems.

Any ideas?
Thanks,
Josh.

jodavies:
OK, I think the issue is that if you arrive at https://github.com/vermaseren/form/blob/master/sources/sort.c#L963 the problem is not that MaxTermSize is too small, but actually you need larger SubSmallSize or SubTermsInSmall etc.

The documentation does describe subfilepatches, sublargepatches, sublargesize though, should multi-stage sorting not be working here?

Thanks,
Josh.

vermaseren:
To my knowledge that ‘subsystem’ works, but indeed, you may have to set those to larger
values than the default.
Much of this was designed before the polyratfun existed and in those days it was hardly
ever used. The polyratfun however makes these parameters much more important.

Jos

jodavies:
Here is a simpler example,
```
#-                             
                               
#: SubTermsInSmall 12K         
#: MaxTermSize 200K              
                               
CFunction f;                   
AutoDeclare Symbol x;          
                               
Local test1 = x;               
Local test2 = f(x);            
                               
Argument f;                    
   Identify x = (x1+...+x10)^7;
EndArgument;                   
Identify x = (x1+...+x10)^7;   
                               
.end                           
```
which gives
```
Time =       0.01 sec    Generated terms =      11440
           test1         Terms in output =      11440
                         Bytes used      =     513660

Time =       0.02 sec    Generated terms =          1
           test2         Terms in output =          1
[...]

Issue #80, opened 2016-04-14
Exhausting disk space when a $-variable doesn't fit in the sub-large buffer


tueda:
With the default setup parameters (Linux x86_64), the following program starts to make a big `xformxxx.sf2` file and seems to enter an infinite loop:

```
S x1,...,x1000;
CF f;
L F = 1;
$x = f((x1+...+x20)^3);
inside $x;
  multiply x1+...+x248;
endinside;
.end
```

If I change `x248` to `x247`, it immediately finishes.

vermaseren:
I guess you are in uncharted territory here.
The $ variables (and function arguments) were never meant for big things.
The $-s use the same buffers as the function arguments for sorting and when you sent these function
buffers into stage4 or worse you may run into things nobody ever tested.
If you really have to do such big things in dollar variables, you better raise the sizes of the various
buffers that are used for this.

Jos

jodavies:
I was looking at this one because it seems similar to several other sorting-related issues that are currently open. It is a stagesort (but not a stage4 sort), it is merging two file patches at the end of the sort.

Eventually it hits a valgrind error, because it reads out of bounds when it hits:
https://github.com/vermaseren/form/blob/0112aa6b21b48616575c6cd205ec00c7797dbc93/sources/sort.c#L2619-L2620
with `*t2=266`. Next iteration `*t1=266`, and after that it looks like it gets stuck in a loop, creating the large file.

The term data in (one of?) the patches gets corrupted somewhere, it looks like.

PR #515, opened 2024-05-09
Make sure SIOsize is >= MaxTermSize

jodavies:
Various buffers in AllocSort depend on the IOsize parameter. At the
main level this is >= MaxTermSize, but this was not the case when
sorting function args, dollar variables.

Fixes the examples in #80 #292

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2277,6 +2277,29 @@ assert succeeded?
 assert result("test", -2) =~ expr("f(0)*g(2)")
 assert result("test", -1) =~ expr("h(2,0)")
 *--#] Issue277 : 
+*--#[ Issue292 :
+#-
+#: MaxTermSize 16K
+#: SubTermsInSmall 800
+#: SubSortIOSize 8K
+
+CFunction f;
+AutoDeclare Symbol x;
+
+Local test = x - f(x);
+.sort
+
+Identify x = (x1+...+x11)^4;
+Argument f;
+	Identify x = (x1+...+x11)^4;
+EndArgument;
+.sort
+
+Identify f(x?) = x;
+Print test;
+.end
+assert result("test") =~ expr("0")
+*--#] Issue292 :
 *--#[ Issue307 :
 * replace_ with nested CFunctions crashes
 S N,j1;
--- a/sources/setfile.c
+++ b/sources/setfile.c
@@ -604,6 +604,8 @@ int AllocSetups(VOID)
 	AM.SMaxFpatches = sp->value;
 	sp = GetSetupPar((UBYTE *)"subsortiosize");
 	AM.SIOsize = sp->value;
+	/* As for IOsize, make sure SIOsize is at least as large as AM.MaxTer. */
+	if ( AM.SIOsize < AM.MaxTer ) { AM.SIOsize = AM.MaxTer; sp->value = AM.SIOsize; }
 	sp = GetSetupPar((UBYTE *)"spectatorsize");
 	AM.SpectatorSize = sp->value;
 /*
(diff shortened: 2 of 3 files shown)


jodavies:
I am unsure about adding a test for #80 : in that case failure means writing out enormous files on github actions...

jodavies:
With 32bits, we can't have a MaxTermSize large enough for this test, it is limited by
https://github.com/vermaseren/form/blob/0112aa6b21b48616575c6cd205ec00c7797dbc93/sources/setfile.c#L437

This test needs to be skipped for 32 bits probably, or I will try to make a version of the test with smaller expressions and buffer sizes that works for 32bit also.

vermaseren:
How about a preprocessor variable that tells the size of a WORD?

jodavies:
Hi Jos,

At least in the test suite it is already easy to disable tests for 32bit word size.

I was actually trying to change the allocation to help with debugging these cases, but got stuck at some point. I want to split the large allocation in AllocSort into lots of smaller ones, so that buffer overruns there will trigger a valgrind error. But it seems that some buffers at least, must be contiguous: the large buffer + extended small buffer, for instance. Are there any others that are used in a combined way?

vermaseren:
Hi Josh,

I am not sure. There are more than 500 calls with Malloc1.
Most are clearly harmless. If a compiler buffer has something, it should also be harmless.
There may be something in the checkpoints, but again, that should not be an issue here.
Also the scratch files should be OK, because a bracketbuffer is not really part of that.

The reason the sort buffer is set up like that because it can be used this way as one giant cache.
But as far as I know that should be all. But many things are more than 30 years ago.
Actually this year it is 40 years ago I started Form. Imagine….

tueda:
> At least in the test suite it is already easy to disable tests for 32bit word size.

It is actually "difficult" because now FORM doesn't give any information about 32/64bit, and that's why by `make check` you are always warned as
```
warning: failed to get the wordsize of '../sources/form'
warning: assuming wordsize = 4
```
It was "easy" in version 4.3.

jodavies:
Sure, but still it is easy to use `#require wordsize == 4` right? Anyway, at least in this case the test is now suitable for 32bit builds.

While investigating this I noticed that the "default" buffer sizes are somehow "not good", since AllocSort alters a lot of them before making the actual allocation. It is a bit hard for the user to know if their specified buffers are actually used in the end. Maybe some warnings could be printed from there (at `allwarnings` level?) for interested users...

tueda:
> Sure, but still it is easy to use `#require wordsize == 4` right?

Right. Indeed, for 32-bit containers, we inform the test suite that the FORM binary uses `wordsize = 2`.

https://github.com/vermaseren/form/blob/0112aa6b21b48616575c6cd205ec00c7797dbc93/.github/workflows/test.yml#L280

> While investigating this I noticed that the "default" buffer sizes are somehow "not good", since AllocSort alters a lot of them before making the actual allocation.  It is a bit hard for the user to know if their specified buffers are actually used in the end. Maybe some warnings could be printed from there (at allwarnings level?) for interested users...

Probably developers also need this. In some test cases, we fix many buffer sizes to reproduce specific conditions. If the way to alter the buffer sizes is changed in future versions (which may be triggered by changing the default settings for other buffers, like [this](https://github.com/vermaseren/form/wiki/Release-Notes-Form-4.2.1#caveats)), then our assumptions about the fixed buffer sizes could easily be invalidated. We need errors when the default or specified parameters are altered to different values.

jodavies:
Errors might be a bit of a pain for regular use though (and like the tests, various real-world scripts will suddenly stop running), but maybe a mode that can be enabled? `On ExactBuffers;` or something.

tueda:
Maybe this PR is ready to merge and indeed has a high priority?

jodavies:
Yes, this one is important. Several of the recently committed tests are effectively implementing this by manually setting `SubSortIOSize` to be able to run. Should we remove those manual settings again, along with this?

tueda:
I think for now we can keep the manual buffer settings in the tests (unless they conflict with the automatic settings).

I will merge this PR as well.

Issue #183, opened 2017-04-28
Crash by transform,mulargs when Sorted function argument too long


tueda:
```
CF f;
Auto S x;
L F = f(<x1+x2+x3+x4>,...,<x7+x8+x9+x10>);
*repeat id f(x1?,x2?,?a) = f(x1*x2,?a);  * error: Sorted function argument...
transform f,mulargs(1,last);  * silent crash
P;
.end
```
terminates without saying the reason:
```
FORM 4.1 (Apr 26 2017, v4.1-20131025-339-gf92e6da) 64-bits  Run: Fri Apr 28 19:25:12 2017

Program terminating at test.frm Line 6 -->
Segmentation fault (core dumped)
```

PR #516, opened 2024-05-10
Change EndSort type for transform statements

### LINKED ISSUE #230: Crash in transform mulargs
Hello,

I am trying to use `Transform f mulargs(1,last)` to efficiently invert a `FactArg` statement. I find either a crash or segfault, depending on some buffer sizes. The issue is triggered by
```
#: MaxTermSize 1M
#: WorkSpace 400M

AutoDeclare Symbol x;
CFunction f;

Local test = f(x1*...*x3502);
.sort

FactArg f;
Transform f mulargs(1,last);
Print +s;
.end
```

To see the crash in `vorm` (all tests with 4.2.0) the number of arguments must be increased to 3504.

Under `valgrind`, for me `vorm` does NOT crash, but yields `f(0)` and gives
```
==30276== Warning: set address range perms: large range [0x1077a040, 0x2f2065c8) (undefined)
==30276== Conditional jump or move depends on uninitialised value(s)
==30276==    at 0x4B7B2F: Generator (proces.c:3072)
==30276==    by 0x4B9473: Generator (proces.c:3931)
==30276==    by 0x4BD86F: InFunction (proces.c:2046)
==30276==    by 0x4B8DED: Generator (proces.c:3759)
==30276==    by 0x4B9473: Generator (proces.c:3931)
==30276==    by 0x4BA9F3: Processor (proces.c:404)
==30276==    by 0x438717: DoExecute (execute.c:838)
==30276==    by 0x44EF96: ExecModule (module.c:274)
==30276==    by 0x4B0A12: PreProcessor (pre.c:962)
==30276==    by 0x4EA969: main (startup.c:1605)
```

Curiously, if I modify the definition of the test expression to, say,
```
CFunction g;
Local test = f(<g(x1)>*...*<g(x8750)>);
```
The code runs in `form`, but not in `vorm`...

Thanks,
Josh.

EDIT: The `SplitArgs/Transform f addargs(1,last)` equivalent of this script appears to work without issues.

jodavies:
Since the transform statements are sorting into the workspace, and might have to merge patches, make them use a par=1 EndSort. This commit makes this change for all transform statements, though only mularg is going to produce an explosion of terms here.

This fixes #230 and causes #183 to crash with a useful error (Term too complex during normalization).

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -1694,7 +1694,22 @@ P;
 .end
 assert succeeded?
 assert result("F") =~ expr("rat(1,ep^2 + 2*ep + 1)")
-*--#] Issue180 : 
+*--#] Issue180 :
+*--#[ Issue183 :
+#: MaxTermSize 16K
+#: SubTermsInSmall 800
+
+CF f;
+Auto S x;
+L F = f(<x1+x2+x3+x4>,...,<x6+x7+x8+x9>);
+*repeat id f(x1?,x2?,?a) = f(x1*x2,?a);  * error: Sorted function argument...
+transform f,mulargs(1,last);  * silent crash
+P;
+.end
+# Runtime errors may freeze ParFORM.
+#pend_if mpi?
+assert runtime_error?("Term too complex during normalization")
+*--#] Issue183 :
 *--#[ Issue185 :
 * Wrong result of content_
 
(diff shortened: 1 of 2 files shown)


jodavies:
@vermaseren this one needs a sanity check, I am not 100% sure of the consequences of this change.

vermaseren:
I think you are right. MulArg was added later. Before this was not a problem, and apparently when MulArg was added this problem was not considered.
And of course there may be future options.

check/fixes.frm
```diff
@@ -1694,7 +1694,25 @@ P;
...
+CF f;
+Auto S x;
+L F = f(<x1+x2+x3+x4>,...,<x6+x7+x8+x9>);
+*repeat id f(x1?,x2?,?a) = f(x1*x2,?a);  * error: Sorted function argument...
+transform f,mulargs(1,last);  * silent crash
+P;
+.end
+# Runtime errors may freeze ParFORM.
+#pend_if mpi?
+assert runtime_error?
```
tueda, line 1711: Here you can write: `assert runtime_error?("Term too complex during normalization")`.

check/fixes.frm
```diff
@@ -2002,6 +2020,28 @@ Symbol x;
...
+Local test = f(x1+...+x11) - (x1+...+x11)^4;
+Identify f(x1?) = f(x1,x1,x1,x1);
+.sort
+
+Transform f mulargs(1,last);
+Identify f(x1?) = x1;
+
+Print +s;
+.end
+assert result("test") =~ expr("0")
```
tueda, line 2043: I think this test should successfully finish: `assert succeeded?`.

jodavies:
It is probably worth mentioning that `mularg` can be slow compared to repeated `id` statements (which are sorting the arguments each repeat).

tueda:
This is ready to merge, right?

PR #490, opened 2024-03-06
Remove the hard limit due to EATTENSOR

tueda:
Try to fix #268.

I'm still not sure why `EATTENSOR` is called so (maybe there was a certain reason [21 years ago](https://github.com/vermaseren/form/commit/b80588f41583a8ebd6e108f7d9a440f1ecfcd969)).

PR #519, opened 2024-05-14
Fix bug in #490

### LINKED ISSUE #523: Tensors don't accept argument field wildcards anymore
The following code crashes with `Illegal substitution of argument field in tensor`:
```form
t t;

l foo = t;

id t(?a) = t(?a);
.end
```
Adding actual index arguments to `t` does not seem to change anything.

`git bisect` tells me that the new behaviour was introduced in 029a64708c9. Since this change is not mentioned in the commit message I assume it was not intentional?

tueda:
This aims to fix the bug in #490 and add related tests.

TODO: code related to `EATTENSOR` in `symmetr.c` is not covered by the tests.

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2727,6 +2727,47 @@ assert stdout =~ exact_pattern(<<'EOF')
    nonempty(14): x y
 EOF
 *--#] Issue486 :
+*--#[ Issue490_1 :
+T t;
+I i1,...,i4;
+L F = t(i1,i3)*t(i2,i4);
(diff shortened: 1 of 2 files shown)


tueda:
I think I still don't get what `EATTENSOR` is, and how to reach the `eattensor` case in `CheckWild()`...

https://github.com/tueda/form/blob/02a58fa01ff58feb5686562d8274238ed9a56763/sources/wildcard.c#L1876

https://coveralls.io/builds/67565397/source?filename=sources%2Fwildcard.c#L1876

PR #520, opened 2024-05-19
Pass the master struct in MasterMerge

jodavies:
CompareTerms needs to be able to change flags in the caller's sort struct (PolyWise, for eg). Using the BHEAD macro here does not pass the right struct.

This leads to polyratfun merging bugs when using tform -w2 (which disables the sortbots).

I think changing the argument of MergeWithFloat won't do anything, but just in case some functionality is added there in the future...


tueda:
It works, so I will merge it. Thanks!

PR #521, opened 2024-05-20
test(check.rb): add #extra_args instruction

tueda:
This is probably the simplest approach to resolve the issue (testing `tform -w2`) discussed in  (https://github.com/vermaseren/form/pull/520#issue-2304774289).

This patch adds a new special instruction `#extra_args` that can be used to add extra command-line arguments to be passed to FORM. For example
```
#extra_args -w2
```
leads to run `tform -w4 -w2 1.frm` for testing TFORM.

Instead of adding `#extra_args`, other options would be:
1. Run all TFORM tests with `-w2` and then with `-w4`.
2. Add 
    ```
    #threads 2,4
    ```
    to run TFORM tests with `-w2` and then with `-w4`.

### DIFF
--- a/check/check.rb
+++ b/check/check.rb
@@ -347,7 +347,7 @@ module FormTest
         nfiles.times do |i|
           ENV["FORM"] = FormTest.cfg.form_cmd
           @filename = "#{i + 1}.frm"
-          execute("#{ulimits}#{FormTest.cfg.form_cmd} #{@filename}")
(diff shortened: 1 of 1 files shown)


tueda:
You are right. Adding `-w2` actually caused many test failures... (I'll create a PR).

tueda:
This patch is not needed for now, but we may revisit it later.

Issue #252, opened 2017-12-13
Memory leaks in nested procedures


tueda:
The first bug of #250 can be reproduced by
```
#procedure foo()
  #procedure bar()
  #endprocedure
#endprocedure

#call foo()
#call foo()
.end
```
Then `valgrind --leak-check=full vorm test.frm` gives
```
==25298== Memcheck, a memory error detector
==25298== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==25298== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==25298== Command: vorm test.frm
==25298==
FORM 4.2.0 (Dec 13 2017, v4.2.0-28-g4493f01) 64-bits  Run: Wed Dec 13 18:25:58 2017
    #procedure foo()
      #procedure bar()
      #endprocedure
    #endprocedure

    #call foo()
    #call foo()
    .end
  0.02 sec out of 0.02 sec
==25298==
==25298== HEAP SUMMARY:
==25298==     in use at exit: 387,475,840 bytes in 125 blocks
==25298==   total heap usage: 646 allocs, 521 frees, 387,529,621 bytes allocated
==25298==
==25298== 8 bytes in 1 blocks are definitely lost in loss record 9 of 120
==25298==    at 0x4C2AB96: malloc (vg_replace_malloc.c:299)
==25298==    by 0x4FFA16: Malloc1 (tools.c:2230)
==25298==    by 0x4FFA5A: strDup1 (tools.c:1816)
==25298==    by 0x4AC6D5: DoProcedure (pre.c:3880)
==25298==    by 0x4AF3D2: PreProInstruction (pre.c:1156)
==25298==    by 0x4AFF3A: PreProcessor (pre.c:931)
==25298==    by 0x4E8F2D: main (startup.c:1607)
==25298==
==25298== 136 bytes in 1 blocks are definitely lost in loss record 59 of 120
==25298==    at 0x4C2AB96: malloc (vg_replace_malloc.c:299)
==25298==    by 0x4FFA16: Malloc1 (tools.c:2230)
==25298==    by 0x4AC0FD: PreLoad (pre.c:4132)
==25298==    by 0x4AC64A: DoProcedure (pre.c:3867)
==25298==    by 0x4AF3D2: PreProInstruction (pre.c:1156)
==25298==    by 0x4AFF3A: PreProcessor (pre.c:931)
==25298==    by 0x4E8F2D: main (startup.c:1607)
==25298==
[...]

PR #524, opened 2024-05-21
Draft: Add color.h test

jodavies:
This test seems trickier than expected. Currently:

`mpirun -np {1,2} parform`: OK
`mpirun -np {3,4} parform`: hang?
`mpirun -np {5,6} parform`: crash

`valgrind vorm`: OK
`valgrind tvorm -w2`: mostly hangs, sometimes finishes in 2s
`valgrind tvorm -w4`: OK, finishes in 2s

The CI sees valgrind errors in `vorm`, `tvorm` that I can't reproduce locally.


jodavies:
I added some print statements to `PF_UnpackRedefinedPreVars` to try to work out what happens. There are many successful redefines, and then we have:
```
0 i = 0
0 trying to redefine ik1 (35) to 1
0    loop j = 0; j < 2
0       AC.pfirstnum[0] (ik1c) (37), index 35
0       AC.pfirstnum[1] (adj) (38), index 35
```
so it fails to find the variable it is trying to redefine in `pfirstnum`. So then it evaluates
```
if ( AC.inputnumbers[j] < inputnumber ) {
```
for `j=2`, causing the `Conditional jump or move depends on uninitialised value(s)`.

So, why is `ik1` no longer in the `AC.pfirstnum` array? Earlier in the program it was there (and had index 35).

tueda:
Thanks for the investigation. I will look into the ParFORM issue. (You know, in programming, the person you were a month ago is a stranger. Then, the person you were more than 10 years ago is...)

By the way, maybe this (the code that gives Valgrind error) should be broken up into small unit tests.

jodavies:
It seems the problem with valgrind and `tvorm -w2` is due to the load balancing. The same issue happens with `-w3`. If I run this test under callgrind, we see that `ThreadsProcessor` makes 100s of millions of calls of `LoadReadjusted` (which is stealing terms from the working thread and distributing them around the idle threads) which also involve locks.

With `w4`, there are only ~400 calls of `LoadReadjusted`.

jodavies:
Once #525 is merged we can rebase this on top and the parform tests will run successfully also.

jodavies:
Here is a simpler way to reproduce the load-balancing issue under valgrind. I cut the expression out of the color test.

This script does basically nothing at all, but on my system `valgrind tform -w2 test-lb.frm` takes about 30 seconds and results in millions of calls to `LoadReadjusted` and its pthread mutex. Using 4 threads or regular form, it finishes in a fraction of a second.

```
#-
#include color.h

Local Q10 =
    + NA*I2R*cR*[cR-cA/2]^3
    - 1/2*NA*I2R*cA*[cR-cA/2]^3
    - 3*NA*I2R*cA*cR*[cR-cA/2]^2
    + 5/2*NA*I2R*cA^2*[cR-cA/2]^2
    + 7/4*NA*I2R*cA^2*cR*[cR-cA/2]
    - 25/12*NA*I2R*cA^3*[cR-cA/2]
    - 1/12*NA*I2R*cA^3*cR
    + 19/48*NA*I2R*cA^4
    + 5*cOlTt(cOlpA1,cOlpA1,cOlpA1,cOlpA1)*[cR-cA/2]
    - 17/6*cOlTt(cOlpA1,cOlpA1,cOlpA1,cOlpA1)*cA
    + 2*cOlTt(cOlpA1,cOlpA1,cOlpA1,N1_?,N2_?)*f(cOlpA1,N1_?,N2_?)*i_
    + cOlTt(cOlpA1,cOlpA1,N1_?,cOlpA1,N2_?)*f(cOlpA1,N1_?,N2_?)*i_
    + cOlTt(cOlpA1,cOlpA1,N1_?,N2_?)*f(cOlpA1,N1_?,N3_?)*f(cOlpA1,N2_?,N3_?)
    - 1/6*cOlTt(N1_?,N2_?,N3_?,N4_?,N5_?)*f(N1_?,N3_?,N6_?)*f(N2_?,N5_?,N7_?)*f(N4_?,N6_?,N7_?)*i_*cA
    + 1/6*cOlTt(N1_?,N2_?,N3_?,N4_?,N5_?)*f(N1_?,N5_?,N6_?)*f(N2_?,N3_?,N7_?)*f(N4_?,N6_?,N7_?)*i_*cA
    - 1/12*cOlTt(N1_?,N2_?,N3_?,N4_?,N5_?)*f(N1_?,N6_?,N7_?)*f(N2_?,N3_?,N6_?)*f(N4_?,N5_?,N7_?)*i_*cA
    - 1/12*cOlTt(N1_?,N2_?,N3_?,N4_?,N5_?)*f(N1_?,N6_?,N7_?)*f(N2_?,N3_?,N7_?)*f(N4_?,N5_?,N6_?)*i_*cA
    - 1/12*cOlTt(N1_?,N2_?,N3_?,N4_?,N5_?)*f(N1_?,N6_?,N7_?)*f(N2_?,N5_?,N6_?)*f(N3_?,N4_?,N7_?)*i_*cA
    - 1/12*cOlTt(N1_?,N2_?,N3_?,N4_?,N5_?)*f(N1_?,N6_?,N7_?)*f(N2_?,N5_?,N7_?)*f(N3_?,N4_?,N6_?)*i_*cA
    - 2/3*cOlTt(N1_?,N2_?,N3_?,N4_?)*f(N1_?,N2_?,N5_?)*f(N3_?,N4_?,N5_?)*cA*[cR-cA/2]
    + 7/18*cOlTt(N1_?,N2_?,N3_?,N4_?)*f(N1_?,N2_?,N5_?)*f(N3_?,N4_?,N5_?)*cA^2
    + 2/3*cOlTt(N1_?,N2_?,N3_?,N4_?)*f(N1_?,N4_?,N5_?)*f(N2_?,N3_?,N5_?)*cA*[cR-cA/2]
    - 7/18*cOlTt(N1_?,N2_?,N3_?,N4_?)*f(N1_?,N4_?,N5_?)*f(N2_?,N3_?,N5_?)*cA^2
   ;
.sort

Print +s;
.end
```

tueda:
Do you eventually want to put `color.frm` and `color.h` here?

The former does not comply with the [rule in README](https://github.com/vermaseren/form/tree/master/check#where-to-add-test-cases) (modulo `user.frm`). The latter is a large file and is originally distributed at the Nikhef site.

Perhaps the rule could be reconsidered and rewritten. The `check.rb` could include a functionality to automatically download libraries as needed.

jodavies:
`color.h` could be fetched by the CI config as `forcer.h` currently is, and the test run in the same way? It doesn't then run locally with a `make check` then, but is still tested on commit.

tueda:
If we want to run the CI tests for `color.frm` in the same way as we do for `forcer.frm` (and it is not necessary for everyone to test `color.frm` locally), then we can configure this in `test.yaml`.

For running Valgrind/coverage jobs for `color.frm` on the CI, currently I'm not sure if it is ready to use, as I forgot if I wrote it in an extendable and general way. Forcer is not compatible with Valgrind, so I didn't need it.

jodavies:
What do you mean, by forcer not being compatible with valgrind?

tueda:
Forcer always hits memory leaks of FORM, if my memory is correct.

jodavies:
OK, I will try to test that. Ideally both the color and forcer tests can pass valgrind too.

tueda:
Here is the issue that is problematic for Forcer with Valgrind: #252

jodavies:
Currently the forcer test (and color, if it moves to the same status) are not part of the coveralls statistic. Perhaps they should also go into that run. At least on my machine, the forcer test running with `vorm` takes only 8s.

jodavies:
@tueda how does this seem? Now "extra" runs both the forcer and color test, and I added them to the coverage run. Further tests which require external libraries or otherwise somehow don't fit in the default tests could also go into "extra".

tueda:
Oh, now I see what you meant, but just to clarify: (1) You don't want to collect `checkpoint` and `multithreaded` into `extra`, do you? And big tests will be put in `extra`. In future, adding many big tests in `extra` in this way may slow down this job and we would need to split it again. (2) Will you add Valgrind check for `extra`? See: https://github.com/vermaseren/form/blob/3262c64592d69c32196b542a3e3c177b45a9545f/.github/workflows/test.yml#L181
That said, this principle might be open to discussion.

tueda:
OK, placing all the extra tests in `extra` is beneficial. I agree.

> Is there a particular reason why the coverage check should only be for the tests which run under valgrind?

It depends on how we define the scope of test coverage. Measuring coverage for Valgrind-validated tests ensures that our tests include verification of the absence of memory errors, at least those detectable by Valgrind (though currently there are `#pend_if` for `valgrind?`, which skip this verification during coverage measurement).

jodavies:
Right, this does make sense; it is good to know the % of code which is both covered and valgrind-error-free. In this case is there a way to fake the valgrind flag, so that the tests which are `#pend_if valgrind?` are also skipped in the lcov run?

Then all test runs could "just run everything", and rely on the `#pend_if` flag to do the skipping. Then we could have, for eg, the color test running with valgrind `vorm` but not `tvorm` for now.

tueda:
> is there a way to fake the valgrind flag

I think I can add a command option for this. `--fake-valgrind` or do you have a better name?

jodavies:
`--fake-valgrind` sounds fine to me. So once that exists, the next steps would be:
+ move multithreaded and checkpoint to "extra"
+ run the "extra" tests also in the valgrind run, with `#pend_if valgrind?` where necessary (eg the forcer test)
+ run the lcov checks with `--fake-valgrind`

More generally, is it possible to have a "build" job which caches the executables, which are then used by all of the checks? Currently all 35 valgrind jobs compile form, as I understand it. Maybe this will improve performance a bit, though I expect some of the harder valgrind tests will still be the bottleneck.

tueda:
> is it possible to have a "build" job which caches the executables, which are then used by all of the checks?

This might be worth a try. But there could be some time overhead for uploading/downloading executables, and hardware or OS versions may not be guaranteed to fully match in different jobs(?) (I've had a bad experience with Google Colab VM, though it is a different vendor.)

jodavies:
OK, then in that case maybe it is fine as it is. At least, the tests only build the binaries that they use so the build is fairly fast.

vermaseren:
You are allowed to put the color.h in the examples. There is nothing proprietary about it.

jodavies:
This is one option indeed. Another could be to host a mirror of the packages on the nikhef website in a new repository in the form-dev organisation

tueda:
> The CI should pass in principle, but it looks like jobs are failing to fetch color.h from nikhef.nl. I am not sure if there are just temporary issues there or there is some rate limiting, in which case we need to fetch the file from elsewhere...

I think this is a "temporary" issue because 6 other jobs successfully accessed the Nikhef server and downloaded `color.h` almost simultaneously (even though some of them didn't need `color.h`; this point can be improved). Once the library is cached, the job works fine.

jodavies:
Which jobs are fetching color.h unnecessarily? It is all conditional on running the “extra” tests isn’t it?

tueda:
Oh, sorry, I was wrong. I saw that `Install library if necessary` ran for all the jobs (even for `default`), but not all the jobs downloaded the `color.h` file...

jodavies:
I also made tests for mincer2 and minceex. The latter is not clean under valgrind though, I’ll try to make a simple reproducing example.

PR #527, opened 2024-05-24
Try to create the sort file at startup.

jodavies:
If a bad path for FORMTMPSORT is specified, FORM doesn't notice until it tries to create a sort file and can't. In principle one can create a script which takes a long time to reach this point.

Try to create the file immediately at startup, and Terminate if this fails.

### DIFF
--- a/sources/startup.c
+++ b/sources/startup.c
@@ -802,6 +802,17 @@ classic:;
 	s[-2] = 'o'; *s = 0;
 	}
 /*
+	Try to create the sort file already, so we can Terminate earlier if this fails.
+*/
+	if ( ( AM.S0->file.handle = CreateFile((char *)AM.S0->file.name) ) < 0 ) {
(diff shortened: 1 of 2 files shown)


sources/startup.c
```diff
@@ -801,6 +801,16 @@ classic:;
...
+/*
+	Try to create the sort file already, so we can Terminate earlier if this fails.
+*/
+	if ( ( AM.S0->file.handle = CreateFile((char *)AM.S0->file.name) ) < 0 ) {
+		MesPrint("Could not create sort file: %s", AM.S0->file.name);
+		Terminate(-1);
+	};
+	/* Reset the file.handle and clean up the test file */
+	AM.S0->file.handle = -1;
+	remove(AM.S0->file.name);
```
tueda, line 813: Is it intentional not to close the file here?

jodavies: No, this was not intentional. I will fix it.

tueda (approved):

PR #529, opened 2024-05-27
Split the large sort allocation into separate allocations

jodavies:
Here I have split the large single allocation in `AllocSort` into individual allocations. Only `Large+Small(+extension)` must be contiguous. The aim is that overrunning any of these buffers will cause a crash and not silent corruption of the program state.

In principle this has a performance penalty, though whether it is measurable or not has to be determined by benchmarking.

The second commit adds warning messages, if any of the buffers are adjusted from their requested values. I have changed some of the default (64bit) allocations, such that they are consistent with FORM's own requirements and no warnings are printed. This is not completely resolved yet: the `sublargesize` is still shifted, and more things are shifted in 32bit mode.


jodavies:
The second point is a bit tricky to resolve, since there is some circular interaction between `largesize` and `IOsize`.

I am not sure of the current code:
https://github.com/vermaseren/form/blob/83e3d4185efca2e5938c665a6df9d67d6d9492ca/sources/setfile.c#L884-L903
`IObuffersize` is in units of WORD (L884). Then comes some code which checks `(LargeSize+SmallEsize)` against `MaxFpatches*IObuffersize` (with a constant offset). If it is too small, the large (or small) buffer are increased in size accordingly. This code runs, basically, if the user has set `sortiosize` significantly larger.

But then we make the same check in reverse, and adjust `IObuffersize` accordingly. This check seems broken though; `IOtry` in units of WORD is compared against `IObuffersize` in bytes. If I fix this comparison, the default setup parameters need more changes to satisfy the constraints.

jodavies:
I have cleaned this up a bit, but now I have a few questions:

In RecalcSetups, the buffer size constraints are not consistent with those in `AllocSort`. `AllocSort` only allocates the internally-determined sizes, so it seems like this doesn't matter in the end. But I am not sure of the purpose of `RecalcSetupts`?
https://github.com/vermaseren/form/blob/83e3d4185efca2e5938c665a6df9d67d6d9492ca/sources/setfile.c#L357

In `AllocSetups` the `ziobuffer` size is set to `IOsize`, but this value is modified inside `AllocSort`. This means that the `ziobuffer` is not the same size as `POsize` (it is smaller, but it is still at least `MaxTermSize`). Could this ever cause a problem?
https://github.com/vermaseren/form/blob/83e3d4185efca2e5938c665a6df9d67d6d9492ca/sources/setfile.c#L567-L592

jodavies:
Another thing I noticed: `AllocSort` allocates the `SplitScratch` array of `TermsInSmall/2` pointers for use in `SplitMerge`. But actually this is never used; `SplitMerge` uses the `SplitScratch` arrays in the `N` struct. The array in the sort struct can be removed.

jodavies:
Here the "changes" to the default buffer sizes imply a change to the manual also.

sources/fsizes.h
```diff
@@ -126,10 +126,10 @@
...
 #define SORTIOSIZE 200000L
 
-#define SSMALLBUFFER 500000L
-#define SSMALLOVERFLOW 800000L
+#define SSMALLBUFFER 2560016L
+#define SSMALLOVERFLOW 3840032L
 #define STERMSSMALL 10000L
-#define SLARGEBUFFER 4000000L
+#define SLARGEBUFFER 26880512L
 #define SMAXPATCHES 64
```
tueda, line 133: Can we round these values to make them a bit cleaner?

tueda:
Do you see any performance regression because of the memory nonlocality?

If this is unclear, maybe we could turn on `SPLITALLOC` only with `DEBUGGING`.

sources/fsizes.h
jodavies: I used the current values since they are not actually a change in the sizes; FORM has effectively been using them all along.

We could use, eg, `2560000L, 3840000L`. `26880000L` for SLARGEBUFFER gets set back up since it is set to `MaxFpatches*((IObuffersize+COMPINC)*sizeof(WORD)+2*AM.MaxTer)  - SmallEsize`.

Changing the default buffer sizes potentially has implications on some of the tests, though the ones that are sensitive to the buffer sizes should be setting them in the test script (and usually the test cases need much smaller-than-default values for a quickly-running test).

jodavies:
I've not measured any performance difference, no. This is required in non-DEBUGGING mode by #537 .

sources/fsizes.h
tueda: OK, then we'll leave these values as they are.

tueda:
For reference, what happens with
```c
SmallEsize = (SmallEsize+15) & (-16L);
```
is not guaranteed prior to C23, which standardizes signed integers are two's complement. That said, we can keep this code, as it's now 2024. Anyway, I don't think anyone is interested in running FORM on exotic systems like [Unisys ClearPath Dorado Servers​](https://stackoverflow.com/a/6972551).

tueda:
Many lines are duplicated for both the `SPLITALLOC` case and the non-`SPLITALLOC` case. Do you want to keep these duplicated lines separate? They could be reorganized as:
```c
#ifndef SPLITALLOC
    LONG allocation = ...
    char *allocated = Malloc1(allocation,"sort buffers");
    #define Malloc1(size, msg) (allocated += (size), (void *)(allocated - (size)))
#endif
    sort = Malloc1(sortsize, "sorting struct");
    ...
#ifndef SPLITALLOC
    #undef Malloc1
#endif
```

jodavies:
I left the old code there for future debugging purposes, but it could also just be removed entirely if it is neater. One can always use git to bisect a potential bug to this change (and assuming I did not make a mistake in the separated allocations, crashes due to this in principle imply bugs elsewhere in the code).

tueda:
> One can always use git to bisect a potential bug to this change

Yes, you are right. We can simply remove the old code, which lowers the code maintenance overhead.

tueda (approved):
Now it looks good to me.

PR #530, opened 2024-05-28
ParFORM: fix #525

### LINKED ISSUE #525: ParFORM redefine bug with more than one expressions
Minimized example from the ParFORM issue in https://github.com/vermaseren/form/pull/524.

`mpirun -np 3 parform` freezes with the following input:
```form
L F1 = 1;
L F2 = 1;
.sort
#define flag "0"
if (expression(F1)) redefine flag "1";
.sort
#if `flag'
.sort
#endif
.end
```

tueda:
This fixes #525 by moving the broadcast of the redefined preprocessor variables to after processing each expression, rather than at the end of the module.

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2841,6 +2841,23 @@ print;
 assert succeeded?
 assert result("F") =~ expr("5000")
 *--#] Issue508 :
+*--#[ Issue525 :
+#:threadbucketsize 5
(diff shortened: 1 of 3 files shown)


jodavies:
If you are happy with this and merge it, I will look again at the color.h test.

tueda:
Rebased. I think this is OK to merge.

PR #531, opened 2024-05-29
Tablebase readonly mode

Herren:
As suggested by @jodavies during the currently ongoing FORM workshop, I added an option
to the tablebase open command, allowing to open  tablebases in read-only mode.
This allows to load tablebases in shared environments where not all users of the tablebases
have write permissions. In case the `addto` command is called on a readonly tablebase file,
an error is thrown and FORM terminates.

The syntax is simply:
`TableBase "tablename.tbl" open, readonly`

Some documentation on this feature is also added to the relevant manual file.


jodavies:
Currently, if you attempt to open a non-existing tablebase in readonly mode, there is no error and the file is created. I think this case should print an error and terminate FORM.

Herren:
Please have a look at the change. It should work now.

Herren:
I attempted to add two tests, but I think I don't entirely understand how the test framework works....

jodavies:
On my system, the parform valgrind error is also an error in form, but not so on the runner?

Anyway, in `GetDbase` in `minos.c` I think you can just `Terminate(-1);` after your error message.

Herren:
Done.

jodavies:
Ah, it seems #573 has broken the 32bit tests.

tueda:
On the master branch, even in 32-bit containers, `FORM_IGNORE_DEPRECATION` is set and CI runs fine. Rebasing might fix the CI failures.

jodavies:
I think this looks good now. Any further comments, @tueda ?

sources/tables.c
```diff
@@ -770,12 +771,23 @@ int CoTBcreate(UBYTE *s)
 int CoTBopen(UBYTE *s)
 {
 	DBASE *d;
-	DUMMYUSE(s);
+	MLONG rw = 1;
+
+	while ( *s == ',' || *s == ' ' || *s == '\t' ) s++;
```
tueda, line 776: This is a matter of taste, but you could use `SkipSpaces` if you want (provided that `"comtool.h"` is included).

sources/tables.c
```diff
@@ -770,12 +771,23 @@ int CoTBcreate(UBYTE *s)
...
+	
+	if ( *s ) {
+		if ( strcmp((char*)s,"readonly") == 0 ) {
```
tueda, line 779: Is it intentional to be case-sensitive? If not, use `StrICmp` instead of `strcmp`.

sources/tables.c
Herren: Done.

sources/tables.c
Herren: Done.

PR #532, opened 2024-05-30
Evaluate pi_ and other constants as floats

fllor:
This PR addresses some issues discussed during the workshop.

- Add Euler and Euler-Mascheroni constants as builtin symbols (`ee_`, `em_`)
- Allow `Evaluate` statement to insert arbitrary precision floating point values for the three builtin constants (`π`, `e`, `γ`). The values are provided (and cached) by MPFR
- Allow symbols as arguments to `Evaluate` statement, e.g. `Evaluate pi_`
- Test cases are included for various expressions, such as `pi_*sqrt_(3)`, `sqrt_(pi_)`, `pi_*ee_`

Furthermore, this PR fixes a bug in the printing of floating point numbers, that could cause exponents to be truncated and memory errors that would be lead to warnings in valgrind.


fllor:
It seems the failure of the 32-bit tests might be due to a bug somewhere in the floating point engine. The program
```
#StartFloat 64
    
Local X = 2/3;
Local Y = sqrt_(3);
Local Z = pi_;
    
ToFloat;
Evaluate;
    
Print;
.end
```
gives me
```
X = 6.66666666666666666667e-01;
Y = 1.73205080756887729352e+00;
Z = 3.14159265358979323846e+00;
```
on my laptop and 
```
X = 6.66666666666666666666666666667e-;
Y = 1.7320508075688772935274463415e+0;
Z = 3.14159265358979323846264338328e+;
```
on the 32-bit containers.

For some reason, floats are evaluated to a different precision.
Also, note that `Y` in the first case is not rounded correctly. In the second case the exponents are not printed properly.

I am not sure if this is a bug in Form or GMP/MPFR. The internal representations are already different in evaluate.c:668.

jodavies:
There is a known problem with the truncation of trailing zeros and messing up the exponent (see #492) which looks to be also the cause of the valgrind failures in your tests. So that needs to be sorted out but I didn't re-check it yet.

Jos said in his slides that float mode is not expected to work properly in 32bit mode, and it seems that it will be removed/declared "unsupported" for v5 anyway.

Once the valgrind errors are resolved you could also move your tests to features.frm rather than user-tests.frm.

fllor:
Printing of exponents and valgrind errors should be fixed now. 32-bit tests will probably still fail.

jodavies:
@fllor I can't see the logs for the failed tests any more, but I expect you just need to add `#require wordsize >= 4` to make sure they don't run for 32bit FORM. See for eg, some of the tests in fixes.frm which also have this.

tueda:
Maybe use `#pend_if wordsize == 2` instead of `#require`?

jodavies:
Could you squash this into a "feature" and "tests" commit? I think it should be good for merge then.

PR #535, opened 2024-06-05
Fix WorkSpace issues with expression optimization

jodavies:
Here are two changes to make the optimizer easier to use.

The first is to dynamically allocate memory in `Horner_tree`, if the `WorkSpace` does not have enough space. There is no particular reason that the workspace should be used for the sort, here, other than that it exists, and might have capacity.

The second, is to re-use the same `WorkPointer` for every call of `Generator` when the optimized expression is placed into FORM's output. `Generator` modifies `AT.WorkPointer` internally, so each term which was added here was placed further and further into the buffer, unnecessarily.

In principle, if you arrive in `generate_expression` with a very-nearly full WorkSpace there could still be a crash. But this should be rare (?) .

These changes pass `pySecDec`'s test suite.

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2943,3 +2943,18 @@ Local F = rat(f,1);
 #pend_if mpi?
 assert runtime_error?("ERROR: polynomials and polyratfuns must contain symbols only")
 *--#] Issue567_3f :
+*--#[ PullReq535 :
+* This test requires more than the specified 50K workspace.
+#:maxtermsize 200
+#:workspace 50000
+S x1,...,x19;
+L F = (x1+...+x19)^4;
+Format O1;
+.sort
+#optimize F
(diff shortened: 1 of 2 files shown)


tueda:
For the first commit, we can add a test, for example:
```form
#:maxtermsize 200
#:workspace 50000
S x1,...,x19;
L F = (x1+...+x19)^4;
Format O1;
.sort
#optimize F
L G = `optimvalue_';
P G;
.end
assert succeeded?
assert result("G") =~ expr("389")
```
which fails for the master branch.

I am not 100% sure we can reuse the workspace in the second commit, but the absence of test failures suggests that it is working.

tueda (approved):

tueda:
The last commit (for `check/fixes.frm`) has a conflict but should be resolved easily.

PR #537, opened 2024-06-07
Optionally reallocate large+small buffer after each module

jodavies:
This is a test of a request from @msloechner , which makes FORM's resident set size better track its actual memory usage. This comes on top of the split allocations branch, since it only deals with the largest buffer. Since FORM allocates its buffers and keeps them for the whole run, the apparent memory usage remains constant after a "peak". If the OS is under memory pressure it will swap pages, but they might be pages that FORM actually isn't using any more.

It seems there are two ways to do it: free and re-allocate the buffer, or use `madvise` to specify `MADV_DONTNEED`. Both have the same effect and have about the same performance, but there is an impact of more than 10% compared to doing nothing. This shouldn't be done by default, and maybe never "every module". (However, the longer each module takes, with lots of disk access during sorting, etc, the less the performance impact will be).

The options could be, a preprocessor statement like `#reallocatesort` which must be specified when the user really wants this (say, after a heavy module) or maybe just `On reallocatesort;` to enable it for every module which follows. Any thoughts?

@msloechner , can you test this and see if it helps your multiple concurrent jobs scenario? I could also see it being useful when running FORM on cluster machines which kill jobs based on RSS: both of these commits have a lower peak RSS in my tests compared to the original behaviour.

Here is a small MINCER performance test for `orig` (unmodified form), `split` (checking the split allocations have no impact) `realloc` (the first commit) and `madv` (the second):
```
Benchmark 1: nice -n -10 ../bin/tform-test-SR-orig -w12 calcdia.frm > calcdia.log1
  Time (mean ± σ):     16.686 s ±  0.216 s    [User: 152.647 s, System: 1.269 s]
  Range (min … max):   16.252 s … 16.909 s    8 runs

Benchmark 2: nice -n -10 ../bin/tform-test-SR-split -w12 calcdia.frm > calcdia.log2
  Time (mean ± σ):     16.718 s ±  0.320 s    [User: 152.869 s, System: 1.197 s]
  Range (min … max):   16.166 s … 17.113 s    8 runs

Benchmark 3: nice -n -10 ../bin/tform-test-SR-realloc -w12 calcdia.frm > calcdia.log3
  Time (mean ± σ):     18.932 s ±  0.154 s    [User: 156.316 s, System: 6.020 s]
  Range (min … max):   18.730 s … 19.200 s    8 runs

Benchmark 4: nice -n -10 ../bin/tform-test-SR-madv -w12 calcdia.frm > calcdia.log4
  Time (mean ± σ):     18.942 s ±  0.248 s    [User: 157.591 s, System: 5.870 s]
  Range (min … max):   18.585 s … 19.308 s    8 runs

Summary
  nice -n -10 ../bin/tform-test-SR-orig -w12 calcdia.frm > calcdia.log1 ran
    1.00 ± 0.02 times faster than nice -n -10 ../bin/tform-test-SR-split -w12 calcdia.frm > calcdia.log2
    1.13 ± 0.02 times faster than nice -n -10 ../bin/tform-test-SR-realloc -w12 calcdia.frm > calcdia.log3
    1.14 ± 0.02 times faster than nice -n -10 ../bin/tform-test-SR-madv -w12 calcdia.frm > calcdia.log4
```
Here is the RSS profile of this test:
rss

### DIFF
--- a/sources/structs.h
+++ b/sources/structs.h
@@ -1867,6 +1867,7 @@ struct C_const {
     int     MemDebugFlag;          /* Only used when MALLOCDEBUG in tools.c */
     int     OldGCDflag;
     int     WTimeStatsFlag;
+    int     SortReallocateFlag;    /* Controls reallocation of large+small buffer at module end */
 	int     doloopstacksize;
 	int     dolooplevel;
     int     CheckpointFlag;        /**< Tells preprocessor whether checkpoint code must executed.
--- a/sources/structs.h
+++ b/sources/structs.h
@@ -1867,7 +1867,10 @@ struct C_const {
     int     MemDebugFlag;          /* Only used when MALLOCDEBUG in tools.c */
     int     OldGCDflag;
     int     WTimeStatsFlag;
-    int     SortReallocateFlag;    /* Controls reallocation of large+small buffer at module end */
+    int     SortReallocateFlag;    /* Controls reallocation of large+small buffer at module end.
+                                        0 : Off
+                                        1 : On, every module (set by On sortreallocate;)
+                                        2 : On, single module (set by #sortreallocate) */
 	int     doloopstacksize;
 	int     dolooplevel;
     int     CheckpointFlag;        /**< Tells preprocessor whether checkpoint code must executed.
--- a/sources/threads.c
+++ b/sources/threads.c
@@ -4769,6 +4769,75 @@ int IniSortBlocks(int numworkers)
 
 /*
   	#] IniSortBlocks : 
+  	#[ UpdateSortBlocks :
+*/
+ 
+/**
+ *	A version of IniSortBlocks which only updates the pointers in the master's
+ *	buffer, to be used after reallocation of that buffer.
+ */
+int UpdateSortBlocks(int numworkers)
+{
+	ALLPRIVATES *B;
+	SORTING *S;
+	LONG totalsize, workersize, blocksize, numberofterms;
+	int maxter, id, j;
+	int numberofblocks = NUMBEROFBLOCKSINSORT, numparts;
+	WORD *w;
+
+	if ( numworkers == 0 ) return(0);
+
+#ifdef WITHSORTBOTS
+	if ( numworkers > 2 ) {
+		numparts = 2*numworkers - 2;
+		numberofblocks = numberofblocks/2;
+	}
+	else {
+		numparts = numworkers;
+	}
+#else
+	numparts = numworkers;
+#endif
+	S = AM.S0;
+	totalsize = S->LargeSize + S->SmallEsize;
+	workersize = totalsize / numparts;
+	maxter = AM.MaxTer/sizeof(WORD);
+	blocksize = ( workersize - maxter )/numberofblocks;
+	numberofterms = blocksize / maxter;
+	if ( numberofterms < MINIMUMNUMBEROFTERMS ) {
+/*
+		This should have been taken care of in RecalcSetups.
+*/
+		MesPrint("We have a problem with the size of the blocks in UpdateSortBlocks");
+		Terminate(-1);
+	}
+/*
+	Layout:  For each worker
+				block 0: size is maxter WORDS
+				numberofblocks blocks of size blocksize WORDS
+*/
+	w = S->lBuffer;
+	if ( w == 0 ) w = S->sBuffer;
+	for ( id = 1; id <= numparts; id++ ) {
+		B = AB[id];
+		AT.SB.MasterFill[0] = AT.SB.MasterStart[0] = w;
+		w += maxter;
+		AT.SB.MasterStop[0] = w;
+		for ( j = 1; j <= numberofblocks; j++ ) {
+			AT.SB.MasterFill[j] = AT.SB.MasterStart[j] = w;
+			w += blocksize;
+			AT.SB.MasterStop[j] = w;
+		}
+	}
+	if ( w > S->sTop2 ) {
+		MesPrint("Counting problem in UpdateSortBlocks");
+		Terminate(-1);
+	}
+	return(0);
+}
+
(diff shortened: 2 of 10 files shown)


msloechner:
Thanks a lot, @jodavies . This looks indeed promising. I will look into it when I find a bit of time.

jodavies:
In case it is useful to see how your own scripts behave, I logged the RSS during the run with something like
```
../bin/tform-test-SR-split -w8 calcdia.frm > calcdia.log1 &
pid=$!
while kill -0 $pid; do
   ps --pid $pid -o rss=
   sleep 0.2
done > tform-rss-orig.dat
```

jodavies:
I added an On/Off switch for this, I think it is better than just a preprocessor instruction since the option can easily be enabled for multiple modules (nice when calling procedures etc) and can still be used for a "single module" (by turning it off again in the next).

I think the "free and reallocate" version is simpler in that we don't need to worry about aligned memory allocations and whether or not this works in Windows.

msloechner:
I've run an example for my use case, it looks quite suitable: I don't see any significant loss in performance in terms of CPU-time in my case, while benefiting from the reduction in resident set size.
For me this improvement will be mainly of importance in (single-threaded) form, where the time spent in single modules greatly exceeds the time needed for sorting and allocating memory, but which do not benefit from tform parallelisation too much. But I see the benefit also for tform as soon as it's able to combine a lot of single terms in a .sort.
RSS

msloechner:
@jodavies Maybe it would be great to also have a preprocessor command #sortreallocate to be active only for a single module. Could you possibly add this?

jodavies:
Do you strongly prefer that to doing
```
On sortreallocate;
...
.sort
Off sortreallocate;
```
?

Of course it is straightforward to have both options, so I can add a `#sortreallocate` too.

msloechner:
I think it would be a nice feature when you're dynamically generating code with the preprocessor and happen to not have access to the previous .sort statement (but don't want to sort right now). What do you think?

jodavies:
In trying to test this further, it seems that actually it does not work in `tform`. The `SortBlocks` contain pointers into the master's `lBuffer` (set in `IniSortBlocks`) which need to be updated. I have no idea how my benchmarks worked before.

@msloechner if you happen to have your test code still around, are you by chance able to check if you used the reallocation mode or the (now deleted) madvise version?

msloechner:
I just tried to run with the tform binaries I compiled in June for buffer reallocation. The tform buffer reallocation immediately aborts with a segfault.
Regarding the madvise version, I can't find the binary on my disk anymore and it seems like you removed any trace of it from github, so I'm no longer able to test it. But it must have been madvise that I got working back in the days with tform.

jodavies:
Rebased on master (which now has the separated sort allocation). Merged the tform fix into the first commit, to avoid a broken commit in the master history.

jodavies:
This branch just force-enables the reallocation for every module, just to see how it goes on the CI: https://github.com/jodavies/form/tree/sort-realloc-ci-test

The test for Issue508 fails, where we have an 8G ulimit, and a reallocation fails. Outside of valgrind it runs OK with the ulimit, so I think we can ignore this.

sources/structs.h
```diff
@@ -1867,6 +1867,7 @@ struct C_const {
     int     MemDebugFlag;          /* Only used when MALLOCDEBUG in tools.c */
     int     OldGCDflag;
     int     WTimeStatsFlag;
+    int     SortReallocateFlag;    /* Controls reallocation of large+small buffer at module end */
```
tueda, line 1870: Can you write the meaning of `SortReallocateFlag` (because it takes 0, 1, 2 if I correctly understand; 0: off, 1: on, 2: temporarily on)? This will be like the comment for `CheckpointFlag`, though VSCode fails to display such multiline comments: https://github.com/microsoft/vscode-cpptools/issues/5811.

tueda:
Maybe I'll add a minor review comment to this PR today, but I'll probably review the core part tomorrow. (Right now, I'm hitting some minor VSCode C++ extension issues...)

sources/threads.c
```diff
@@ -4769,6 +4769,76 @@ int IniSortBlocks(int numworkers)
...
 /*
   	#] IniSortBlocks : 
+  	#[ UpdateSortBlocks :
+*/
+ 
+/**
+ *	A version of IniSortBlocks which only updates the pointers in the master's
+ *	buffer, to be used after reallocation of that buffer.
+ */
+
```
tueda, line 4779: Doxygen is happy with blank lines between a function and its corresponding comment (but #579 for `threads.c`), while VSCode doesn't seem to like them. Personally, I also prefer not to put blank lines between them, as it looks tidier.

jodavies: Many of the functions in threads.c have a blank line, though. We can remove them all, if you prefer?

tueda: Yes, but removing all such blank lines (maybe also in other files) would be handled in a separate PR. For now, we can focus on the newly added function.

tueda:
The code looks fine.

But why did you choose `#reallocatesort` instead of `ModuleOption reallocatesort` for reallocation at the end of the module? To me, `#reallocatesort` sounds like an immediate reallocation (which would be fine during preprocessing).

jodavies:
Not really any reason I suppose, it would also fit nicely as a moduleoption alongside "inparallel" and "noparallel", i.e. options which don't really have anything to do with the processing of the actual terms (like "local" etc) but more to do with the high-level operation of the module.

I am not opposed to adding a moduleoption (and removing the pre-proc command?) if you prefer.

msloechner:
@tueda: Very sorry for answering only now. I would prefer the first option, because the sortreallocation may only take place at a .sort or .store, and in Form statements like ModuleOption become effective at the end of the module.

Issue #540, opened 2024-06-16
Zstd compression for sort files


jodavies:
I ran a few tests, using @magv 's suggestion of using https://github.com/facebook/zstd/tree/dev/zlibWrapper to use Zstd while making almost no changes to FORM's code. You need to include the wrapper's header instead of `zlib.h`, and call a function to enable the use of zstd.

For the test I link against a library as compiled by https://github.com/magv/hepware . We could do this, or alternatively compile zstd in with FORM (by setting it up as a git submodule or by directly including the code in this repository -- it is 3-clause BSD, I think that is allowed).

My first test was a simple program which generates lots of terms in a single module which all cancel in the final sort; no scratch files are created, only sort files. This runs a few percent faster with zstd, and uses a few percent less disk space.

I also ran a mincer test (an N=11 moment of a single diagram for "graviton-exchange DIS" -- this is an old benchmark I have lying around from Andreas). This uses ~12GB of disk (mostly the uncompressed scratch files) but runs 8% faster.

All these tests ran in tmpfs, so disk performance was never a bottleneck. On a slow disk the better compression ratio (say, 5%?) will help as well.

This seems like an easy improvement -- we just need to decide what the best way to include it in the build is. Any thoughts?

jodavies:
Yes, that looks good. At least for me, using the `#define` before the header include to enable zstd doesn't work properly. But we can call `ZWRAP_useZSTDcompression(1);` (or `ZWRAP_useZSTDcompression(0);`) in the code somewhere. Using this we can just add a sub-key like `On compress, zstd;`.

I would say that if compiled with zstd support, use of zstd should be the default. But then one can specify `On compress, gzip;` for easy benchmarking.

Do you want to implement the On/Off statement also, or I can do it and make a PR into your branch?

tueda:
Please go ahead. `On compress, zstd` sounds good. My implementation is experimental, so feel free to use any part of it in your branch if you find it useful.

jodavies:
Looking through the code for where I need to add things, I noticed that there is https://www.nikhef.nl/~form/maindir/documentation/reference/online/online.html#SECTION008220000000000000000 . Either this needs to also have the zstd option, or we can remove this form FORM 5.

tueda:
> Either this needs to also have the zstd option, or we can remove this form FORM 5.

The manual doesn't mention `Compress gzip`, though it works while `Compress zstd` gives

```
[...]

PR #541, opened 2024-06-19
Zstd compression for sort files

tueda:
This PR implements #540.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -75,7 +75,7 @@ jobs:
         if: runner.os == 'Linux'
         uses: awalsh128/cache-apt-pkgs-action@v1
         with:
-          packages: libmpfr-dev
+          packages: libmpfr-dev libzstd-dev
           version: 1.0
 
       - name: Install dependencies (Windows)
@@ -103,7 +103,7 @@ jobs:
 
       - name: Configure
         run: |
-          opts='--disable-dependency-tracking --disable-scalar --disable-threaded --disable-native --enable-static-link --with-gmp --with-zlib'
+          opts='--disable-dependency-tracking --disable-scalar --disable-threaded --disable-native --enable-static-link --with-gmp --with-zlib --with-zstd'
           case ${{ matrix.bin }} in
             form)  opts="$opts --enable-scalar";;
             tform) opts="$opts --enable-threaded";;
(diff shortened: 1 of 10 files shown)


tueda:
If I remember correctly, it should work. But I don't have a good benchmark showing the performance advantages of zstd compression. (This may just be because all my machines currently use SSDs, not HDDs.)

Do you have a good benchmark results for this? If you are happy with this implementation, I think it can be merged (but the "`Fix previous" commit can be squashed into the previous one).

I'll rebase it.

jodavies:
I will take another look at some benchmarks. At least in #540 I claimed some improvements.

tueda:
zlibWrapper uses the quote form of the `#include` directive for `zstd.h`:

https://github.com/facebook/zstd/blob/794ea1b0afca0f020f4e57b6732332231fb23c70/zlibWrapper/zstd_zlibwrapper.c#L28

For some reason, `CPATH` doesn't work for the quote form while `CPPFLAGS -I` does. I'm confused because the GCC documentation says they have [the same effect](https://gcc.gnu.org/onlinedocs/gcc-9.5.0/cpp/Environment-Variables.html)...

jodavies:
This also needs adding to the manual. Additionally, the current manual claims "On compress;` is the default, but rather `On compress,gzip;` is the actual default.

jodavies:
This is a good benchmark I think. It makes a large expression (but it fits in ScratchSize) and then pointlessly reads and writes it a bunch of times. So a large % of the time is spent in the compression routines.
```
#-

* Keep expressions in memory. We want to test only the sort files.
#: ScratchSize 3G
* Write out patches more regularly during sorting
#: TermsInSmall 10K

Off statistics;
Off threadstats;
`COMPR'

#define NTERMS "{2^20}"
#define BLOCKS "16"
#define ITER "10"

Symbol n,m,x,y;
CFunction f,g,h,sum;

Local test =
	#do b = 1,`BLOCKS'
		+ f(`b') * sum(x,1,{`NTERMS'/`BLOCKS'},g(x))
	#enddo
	;
.sort
Identify sum(?a) = sum_(?a);
* Make the terms larger.
Identify f(n?)*g(m?) = f(n)*g(m) * h((n*x+m*y)^25);

* Pointlessly read and sort and write the terms
#do i = 1,`ITER'
	#message i = `i' / `ITER'
	.sort
#enddo

.end
```

I run it with FORMTMP on a tmpfs, so the speed of the storage is taken out of the equation. Then I see:
```
Benchmark 1: Master branch, default ( = On compress,gzip;)
  Time (mean ± σ):     120.475 s ±  5.269 s    [User: 568.036 s, System: 7.564 s]
  Range (min … max):   114.838 s … 125.275 s    3 runs

Benchmark 2: Master branch, Off compress;
  Time (mean ± σ):     45.697 s ±  0.583 s    [User: 212.365 s, System: 22.688 s]
  Range (min … max):   45.110 s … 46.276 s    3 runs

Benchmark 3: zstd branch, default (= On compress,zstd;)
  Time (mean ± σ):     112.249 s ±  2.962 s    [User: 501.964 s, System: 7.733 s]
  Range (min … max):   108.968 s … 114.724 s    3 runs

Benchmark 4: zstd branch, On compress,gzip;)
  Time (mean ± σ):     121.376 s ±  2.576 s    [User: 566.287 s, System: 7.538 s]
  Range (min … max):   118.985 s … 124.105 s    3 runs
```
The peak size of the sort files 741M, 2.5G, 701M and 741M, respectively. That Benchmark 1 == Benchmark 4 means that using gzip via the wrapper does not incur a penalty.

Benchmark 2 is very fast since we do no compression at all, but if FORMTMP is on a real storage device this slows down a lot.

So Zstd is both a bit faster, and compresses a bit better, in this test.

I'll run a test with a bit more statistics but it already looks reasonable.

Could you rebase the fix commit in your branch, and here is a suggestion for the manual which you can cherry-pick if you are happy: https://github.com/jodavies/form/commit/7071ccea60b340be148744c3cff280104fc4f81d

tueda:
Thanks! The results look nice.

Maybe it would be useful to put this benchmark in the repository? We could simply put it into some subdirectory `bench/` or `benchmarks/`. Alternatively, we could use the test suite framework, putting it to `check/bench/` with an additional test fold and simple assertions, then `./check/check.rb /path/to/form -C bench thisbench.frm` should run it.

I have cherry-picked the patch for the manual and squashed the "Fix previous" commit into the previous one.

jodavies:
I added the benchmark in check/benchmarks, and also a "mini" version in the usual tests. The idea is to ensure there is a test in there which creates sort files (very likely there is, but this guarantees it...).

Under `valgrind tform` the test is surprisingly slow, it suffers from the same load balancing issue as the color test. I am not sure about the parform failure.

tueda:
Rebased (conflicts resolved).

jodavies:
One thing: this works fine when loading tablebases which have been created with "zlib form" since the wrapper detects zlib data and uses the right decompression function. But not the other way around: tablebases created with zstd enabled do not work in "zlib form".

The easiest solution would be to disable and renable the wrapper around calls to compress in minos.c.

PR #570, opened 2024-11-06
Check for bad terms in fast notation in get_variables

### LINKED ISSUE #567: Vector in polyratfun vanishes
In trying to compare OPITeR to other tensor reduction routines I hit the following (due to an autodeclare, I did not notice that `q22` was taken as a vector):

```
#-

Vector q22;
*Symbol q22;

Vector q2;
Symbol D;
CFunction rat;
PolyRatFun rat;

Local test = q2.q2*rat(-1,D - 2);

Identify q2.q2 = rat(q22,1);

Print +s;
.end
```
produces
```
test =
    + rat(0,1)
   ;
```
which then vanishes after another sort.

Putting a scalar product into rat of course crashes like
```
ERROR: polynomials and polyratfuns must contain symbols only
```

This goes at least as far back as FORM 4.1, it is a bit surprising that no one encountered this before.

jodavies:
These created numerators of 0. Fixes #567

### DIFF
--- a/sources/poly.cc
+++ b/sources/poly.cc
@@ -2408,7 +2408,14 @@ void poly::get_variables (PHEAD vector<WORD *> es, bool with_arghead, bool sort_
 				degrees.push_back(1);
 			}
 		}
-		else {		
+		// JD: Here we need to check for non-symbol/number terms in fast notation.
+		else if (*e < 0) {
+			MLOCK(ErrorMessageLock);
+			MesPrint ((char*)"ERROR: polynomials and polyratfuns must contain symbols only");
+			MUNLOCK(ErrorMessageLock);
+			Terminate(1);
+		}
+		else {
 			for (int i=with_arghead ? ARGHEAD : 0; with_arghead ? i<e[0] : e[i]!=0; i+=e[i]) {
 				if (i+1<i+e[i]-ABS(e[i+e[i]-1]) && e[i+1]!=SYMBOL) {
 					MLOCK(ErrorMessageLock);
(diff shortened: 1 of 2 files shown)


tueda:
Should this check be in `poly::argument_to_poly()` or `poly::get_variables()`?

jodavies:
You are right, this check could equally go in `get_variables` and makes more sense there. I will move it.

sources/poly.cc
```diff
@@ -2408,7 +2408,14 @@ void poly::get_variables (PHEAD vector<WORD *> es, bool with_arghead, bool sort_
 				degrees.push_back(1);
 			}
 		}
-		else {		
+		// JD: Here we need to check for non-symbol/number terms in fast notation.
+		else if (*e < 0 && *e != -SNUMBER && *e != -SYMBOL) {
```
tueda, line 2412: `*e` cannot be `-SNUMBER/-SYMBOL` because it is in the else-if-clause.

jodavies: Yes you are right. It redundant conditions might well be optimized away but we might as well remove them. It is clear from the comment what is happening.

tueda (approved):
Now the code looks fine.

Optionally, the code for triggering the error could be reused with `goto`, like:
```c++
		else if (*e < 0) {
term_too_complicated_error:
			MLOCK(ErrorMessageLock);
			MesPrint("ERROR: polynomials and polyratfuns must contain symbols only");
			MUNLOCK(ErrorMessageLock);
			Terminate(1);
		}
		else {
			for (int i=with_arghead ? ARGHEAD : 0; with_arghead ? i<e[0] : e[i]!=0; i+=e[i]) {
				if (i+1<i+e[i]-ABS(e[i+e[i]-1]) && e[i+1]!=SYMBOL) {
					goto term_too_complicated_error;
				}
```

jodavies:
While here it would probably be fine, other areas of the code have been cumbersome to debug because you have to determine which of many `goto` have actually caused a termination. In my opinion it could be beneficial to actually remove this pattern from the code entirely, but that requires some effort also...

vermaseren:
How about giving Terminate a second parameter that indicates which call it was?
Then one can change things easily, using grep.

jodavies:
When this PR is done https://github.com/vermaseren/form/pull/526 , `Terminate` is replaced with a macro which prints the file and line information of every terminate call.

But still if you have several `goto` which land at a `Terminate`, you can't know which `goto` caused the crash without stepping through in a debugger (hard if the crash is in a long running script with no useful way to set a breakpoint) or putting print statements at each of the `goto`.

Issue #475, opened 2024-02-23
[Meta-issue] Fix the FORM5 regressions


tueda:
This is the meta issue to track regressions introduced in 5.0.0-beta. To be updated occasionally.

- #7
- #474

tueda:
I will start merging postponed PRs that are complete and likely to have no or negligible side effects such as #477, #482 and #490. They will be listed in this issue as those that could potentially introduce some regressions (albeit unlikely). In this way, test users can access versions incorporating the latest PR merges, with awareness of potential regressions.

jodavies:
Before FORM 5, I would like to resolve at least the following. For some of these I know roughly what the problem is, but would benefit from a discussion with Jos to check that I really understand things. The workshop in May would be a good place to do this I expect.

#230 
#292 
#420 and #484
#468

Maybe we should also start to work through the open issues and close + tag as "question" or "wontfix" where appropriate, to clean up the issue tracker a bit.

tueda:
Do we need another 5.0 beta version (and a 4.3 version with backported bug fixes) before the workshop? This could serve as a good starting point for discussions at the workshop.

If the workshop successfully resolves the issues you mentioned, we could then proceed with another (beta) release afterwards.

Related to the tags:
The `question` and `good first issue` tags are abused in ways that differ from the [GitHub defaults](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#about-default-labels): the former had been used for "questions by users" (though I haven't put `question` tag recently because of this difference) and an example of the latter is https://github.com/vermaseren/form/issues/300. We might consider creating new tags with more specific meanings. At least, adding a `documentation` tag (as the default meaning) would be good.

jodavies:
I am not sure a specific beta "release" is necessary for the workshop. The current master branch is in reasonable shape to work from (diagram generator issues aside), and the results of the workshop can inform what a final beta of FORM 5 should look like before a testing period and final release.

We should also look through the issues labelled as "enhancement" and think about what would be good to include in FORM 5. (Personally I would like to look at #436 at least).

Since "question" has been used previously to tag questions about the use of FORM, my opinion would be to just keep that convention now.

jodavies:
Here is a bit of a summary of some discussions at the workshop.

**Diagrams/topologies:**
 - The new diagrams_ should have all of the functionality of the old topologies_, which
   is there for backwards compatibility. But it could also be removed from FORM5. Then
   the open bugs regarding this are "resolved".
[...]

PR #573, opened 2024-11-13
Run-time deprecation warning for native Windows, 32-bit systems, ParFORM and Checkpoint

tueda:
This is an implementation of deprecation messages like those in  https://github.com/vermaseren/form/issues/475#issuecomment-2146949223 but **at runtime startup** rather than at build time, for Windows, 32bit builds and ParFORM.

End users may not build binaries themselves. 32-bit binaries and ParFORM are distributed by various [distributors](https://repology.org/project/form/versions). We can distribute native Windows binaries if https://github.com/vermaseren/form/pull/511 is merged and `Delete Windows binaries` in `deploy.yml` is removed. This patch will cause these end users to see annoying deprecation warnings.

The deprecation warning looks like, for example, for ParFORM:
```
DeprecationWarning: We are considering deprecating the MPI version (ParFORM).
If you would like support to continue, please leave a comment at:

    https://github.com/vermaseren/form/issues/xxxxx

Otherwise, it will be discontinued in the future.
To suppress this warning, use the --ignore-deprecation command line option or
set the environment variable FORM_IGNORE_DEPRECATION=1.
```

This warning can be suppressed by the (undocumented) `--ignore-deprecation` option or the environment variable `FORM_IGNORE_DEPRECATION=1`.

TODO: we need 3 separate issues for users to leave comments regarding the deprecation.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -5,6 +5,9 @@ on:
   pull_request:
   workflow_dispatch:
 
+env:
+  FORM_IGNORE_DEPRECATION: 1
+
 jobs:
   # Generate the tarball distribution, e.g., "form-v4.2.1.tar.gz" for v4.2.1.
   # The tarball will be tested in the following "build-bin" job.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,9 @@ on:
   pull_request:
   workflow_dispatch:
 
+env:
+  FORM_IGNORE_DEPRECATION: 1
+
 jobs:
   # Simple tests on Linux; except the ParFORM case, they probably pass unless
   # the committer has forgotten running "make check".
--- a/sources/checkpoint.c
+++ b/sources/checkpoint.c
@@ -840,6 +840,7 @@ static void print_M()
 	MesPrint("%d", AM.ggShortStatsMax);
 	MesPrint("--MARK 11");
 	MesPrint("%d", AM.FromStdin);
+	MesPrint("%d", AM.IgnoreDeprecation);
 	MesPrint("%%%% END M_const");
 /*	fflush(0); */
 }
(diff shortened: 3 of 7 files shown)


jodavies:
A run-time message seems good, as you say users may just use the distributed binaries rather than compiling themselves. This would be the first "--" option, is it more consistent to use a single "-" like `-pipe`? I have no strong opinion there.

tueda:
I was considering the convention for long-name options in GNU tools (other than GCC), but this makes sense (`-pipe` already exists). I'll switch it to `-ignore-deprecation`.

If you have a better name than `-ignore-deprecation`/`FORM_IGNORE_DEPRECATION`, then let me know.

tueda:
Added a deprecation warning for the checkpoint as well, as discussed during the video call.

tueda:
Now the issue numbers are linked to actual issues.

tueda:
I haven't cherry-picked it into 4.3.

I don't remember why (maybe I simply missed them), but should all of the last arguments of `PADPOSITION` be incremented? So, it should be:
```c
#ifdef WITHFLOAT
#ifdef WITHPTHREADS
	PADPOSITION(17,30,62,84,(sizeof(pthread_rwlock_t)+sizeof(pthread_mutex_t)*2)+2);
#else
	PADPOSITION(17,28,62,84,2);
#endif
#else
#ifdef WITHPTHREADS
	PADPOSITION(17,30,62,84,(sizeof(pthread_rwlock_t)+sizeof(pthread_mutex_t)*2)+2);
#else
	PADPOSITION(17,28,62,84,2);
#endif
#endif
```

The commit 7b48c35f1df09018fda1fbb7f47a784b9a26ae36 incremented only the last argument of the second `PADPOSITION`.

jodavies:
Right, yes. In practice I don't think the explicit padding is required these days but we shouldn't let the numbers get "out of date" in the macros.

tueda:
OK, I'll push the fix. Thanks.

Issue #546, opened 2024-07-05
Add `.D0` to integers > i32 in Fortran optimization output


benruijl:
Fortran code generation writes:

```
& 124199770599398327033900700000.D0 - 2669685476*z2
```

Recent `gfortran` compilers emit `(Real*8): the integer 2669685476 is too large`.

The error likely occurs in `VOID RatToLine(UWORD *a, WORD na)` where no `.D0` is added if the number fits in a WORD, which is a unsigned 32-bit number. However, the `.D0` should also be added for integers larger than 2^31.

tueda:
Maybe, we could add a setup parameter (say, `HugeInteger`) that controls the maximum absolute value of integers without a float suffix for both Fortran and C.

`HugeInteger` must be a BigInt parameter (at least 128 bits or so; 4 WORDs in x86_64) with a default value of 2^31-1 in x86_64.

jodavies:
It could be useful to also add
`Format C,_mysuffix;`
like the Fortran90 mode has, such that one could use https://en.cppreference.com/w/cpp/language/user_literal

jodavies:
This is a cleaned up `RatToLine` which is supposed to produce the same output as the master code, except the floating suffix is added for numbers > 2^31-1.

Adding some logic to set `maxInt` depending on C or Fortran mode, or to a user-supplied value should now be straightforward.

https://github.com/jodavies/form/commit/a61e1a6e1c7a2ca1fea0553d18a19e450abf7b12

PR #584, opened 2024-11-19
Fix printing of literals in fortran modes

jodavies:
This supposed to be a "minimal" fix of #546 , in that it fixes the printing of literals `> 2^31-1` in fortran modes by adding the float suffix. It does not change the C mode, nor introduce any user-configurable parameter to choose a value.

There are two more differences w.r.t master, so in some sense it is not an entirely "minimal" fix. Maybe someone has some comments:
- In `fortran90` mode, the master branch adds a "." unconditionally, also to integers `< 2^31-1`. This PR doesn't do this, it prints them in the same way as `fortran` mode.
- The (undocumented) `pfortran` mode does not work properly in the master branch: it doesn't print the `((one*integer)/integer)` format in all cases ("test32" in the test case, for eg).

Additionally, note that in 32bit FORM, the master branch starts adding floating suffixes at `2^15-1` (and this PR at `2^16-1`).


sources/sch.c
```diff
@@ -338,130 +338,116 @@ VOID RatToLine(UWORD *a, WORD na)
 {
 	GETIDENTITY
 	WORD adenom, anumer;
+	UWORD maxInt;
+
+	if ( AC.OutputMode == CMODE ) {
+		// In C mode, over 2^32-1 should be printed with a float suffix (of course 2^32 fits
+		// in a 32bit unsigned, but we don't want to change FORM's behaviour here).
```
tueda, line 345: What do you mean by "of course 2^32 fits in a 32bit unsigned"? 64bit unsigned?

tueda (approved):
I never fully understood what the old code did, but the new code is understandable!

For now, I think the code looks fine as a minimal fix (except for the comment). But if we want to make sure `PFortran` mode works correctly in future,
`4294967296/4294967297` should be output like `(one*2*2147483647+2)/(one*2*2147483647+3)` instead of broken `((one*4294967296.D0)/4294967297.D0)`.

sources/sch.c
jodavies: I am not sure now, so it is not a very good comment... I think the point was that unlike fortran, C will treat >= 2^32 as 64bit literals so the float suffix could be added at 2^64-1, but we should keep FORM's output the same.

tueda: Right. In Fortran, integer literals without `kind-param` is default integer ([J3/23-007r1](https://j3-fortran.org/doc/year/23/23-007r1.pdf), 7.4.3.1). In C, integer literals without suffixes are extended up to `unsigned long long int` ([N3220](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf), 6.4.4.2, 6).

jodavies:
Rebased and comment improved.

PR #586, opened 2024-11-20
Fix expression source file in GetFirstTerm

### LINKED ISSUE #577: .sort or .sort\n .sort
Guten Morgen!

The following code:

```
g ff = 1;
.sort

#if ( isnumerical(ff) )
  #message HERE!
#endif
.sort

print;
.end
```

Here we have no effect of #if.  If I put another .sort, then it works.

jodavies:
During preprocessing, RevertScratch has not yet been called in the current module execution. So the expressions are sitting in the outfile of the previous module.

Fixes #577 

The final test case does not work in tform. The manual states "If the expression has only been defined in the current module and has not yet been processed (as is the case when referring to the first term in the current expression) the answer will be unspecified. This use is considered illegal, even though it does not generate an error message." so maybe it is acceptable to leave things like that?

There is a still a parform valgrind issue to resolve.


jodavies:
The parform problem is because the worker threads are also running this code. Then `*term` appears to be zero in the workers, so they try to access `term[-1]`:
https://github.com/vermaseren/form/blob/fc829afdf051c817fb53f4d614ba6065db54f8c7/sources/pre.c#L4887

Presumably only the master thread should actually be running this?

jodavies:
OK, then we can skip parform for this test for now.

sources/execute.c
```diff
@@ -1851,12 +1851,13 @@ int GetFirstBracket(WORD *term, int num)
...
 */
 
-int GetFirstTerm(WORD *term, int num)
+int GetFirstTerm(WORD *term, int num, int pre)
 {
 /*
 		Gets the first term of the expression 'num'
 		Puts it in term.
 		Routine should be thread-safe
+		The "pre" arg denotes a pre-processor call (1) or not (0).
```
tueda, line 1860: How about reformatting the comments for this function into Doxygen style now that you've added the `pre` parameter and its explanation?

```c
/*
 		#] GetFirstBracket : 
 		#[ GetFirstTerm :
*/

/**
 * Gets the first term...
 *
 * @param[in] term ...
 * @param[in] num ...
 * @param[in] pre ...
 * @return ...
 */
int GetFirstTerm(WORD *term, int num, int pre)
```

jodavies: Sure, done.

tueda (approved):

tueda:
Looks fine to merge.

PR #588, opened 2024-11-25
Add documentation for UNCHANGED_ and ZERO_

### LINKED ISSUE #585: Checking that a local expression is zero
Dear all, 

I know that it's a stupid question, but I really wasn't able to find an elegant way solve it without complicating
everything too much. So is there a simple way to check if the given local expression is zero during the execution?

Consider for example

```
S a,b; 
L exp1 = a+b;
L exp2 = a*b;
.sort
id a=-b;
.sort
* More code
print;
.end
```
here `exp1` becomes zero after the `id` statement, while `exp2` is not zero. Ideally, I would like to detect `exp1` vanishing and finish the program right here (or maybe jump to the end), instead of going through all the subsequent code instructions.

Cheers,
Vlad

jodavies:
This resolves #585

### DIFF
--- a/check/features.frm
+++ b/check/features.frm
@@ -1187,3 +1187,93 @@ else
   assert runtime_error?('Could not create sort file: bad_path\xformxxx.sor')
 end
 *--#] TempSortDir_windows :
+*--#[ ZeroUnchanged :
+#-
+
+#procedure exprinfo
+	#message Module `CMODULE_':
+	#do e = {`activeexprnames_'}
+		#if `ZERO_`e''
+			#message zero `e': `ZERO_`e''
+		#endif
+		#if `UNCHANGED_`e''
+			#message unchanged `e': `UNCHANGED_`e''
+		#endif
+	#enddo
+	#if `ZERO_'
+		#message All zero: `ZERO_'
+	#endif
+	#if `UNCHANGED_'
+		#message All unchanged: `UNCHANGED_'
+	#endif
+	#message
+#endprocedure
+
+
+Off stats;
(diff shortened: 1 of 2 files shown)


jodavies:
I would like to add a test-case for this, but I don't think everything works quite as expected. Here, test3 becomes 0 in the 5th module, but the 6th module's print has test3 as both zero and unchanged, which is not correct.

```
#-

#procedure exprinfo
	#message Module `CMODULE_':
	#do e = {`activeexprnames_'}
		#if `ZERO_`e''
			#message zero `e': `ZERO_`e''
		#endif
		#if `UNCHANGED_`e''
			#message unchanged `e': `UNCHANGED_`e''
		#endif
	#enddo
	#if `ZERO_'
		#message All zero: `ZERO_'
	#endif
	#if `UNCHANGED_'
		#message All unchanged: `UNCHANGED_'
	#endif
	#message
#endprocedure


Off stats;

Symbol x,y;

Local test1 = x;
Local test2 = y;
Local test3 = 1;
.sort

#call exprinfo
Identify x = 0;
.sort

#call exprinfo
Identify y = 0;
.sort

#call exprinfo
.sort

#call exprinfo
Multiply 0;
.sort

#call exprinfo
Print;
.end
```

jodavies:
Maybe it is OK that `UNCHANGED_test3` is not correct in this instance. In general UNCHANGED is not hugely "reliable" anyway; `Identify f(x?) = f(x);` or `Multiply 1;`, etc etc all count as "changes" in this context.

PR #590, opened 2024-11-27
Better warnings for resized sub-buffers in DEBUGGING mode

jodavies:
Add a function argument to AllocSort to specify which buffers are being allocated. This makes it straightforward to print an appropriate warning message.


tueda (approved):
Maybe, it might be worth converting the comment of `AllocSort` into Doxygen style, but it would be tedious and too verbose to make `@param` for all the parameters...
```c
/*
 		#] WriteSetup : 
 		#[ AllocSort :
*/

/**
 * Routine allocates...
 *
 * @param[in] ...
 * @return ...
 */
SORTING *AllocSort(LONG inLargeSize, ...
```
So, I think the current PR is acceptable.

PR #591, opened 2024-11-27
Multirun

jodavies:
Always use "multirun" mode, to write the PID in the temporary file names. This makes FORM more robust by default, when running multiple jobs with the same FORMTMP/FORMTMPSORT directories.

Additionally use more digits of the PID in the name.

### DIFF
--- a/check/features.frm
+++ b/check/features.frm
@@ -1139,9 +1139,5 @@ assert result("F") =~ expr("0");
 #: TempSortDir bad/path
 Local test = 1;
 .end
-if mpi?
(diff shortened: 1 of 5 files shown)


jodavies:
Rebased.

tueda:
I guess merging this PR into 4.3 won't impact typical users not doing anything special (like checking whether temporary files exist). But to be on the safe side, we could merge it only into 5.0 and skip 4.3 (until someone convinces us it must be included in 4.3 as well).

PR #592, opened 2024-11-28
Write term size in compress buffer in GetOneTerm when reading from fi…

### LINKED ISSUE #214: Mincer crash
While testing various disk/compression settings with mincer, I get a strange crash:
```
Time =      17.38 sec    Generated terms =    4700985
            d65c         Terms in output =    3136630
      Prepare first loop Bytes used      = 1181651608
Program terminating in thread 1 at ACCU2 Line 8 --> 
Floating point exception (core dumped)
```
This is using the mincer test in the speedtest dir. I can provide the files if you don't have them. The crash happens for `#define POW "7"` and higher, but ONLY if compression is switched off. I comment out the `on,compress,gzip,6;` line in `treatgzgz.prc` and add `Off compress;` at the top of `calcdia.frm`.

It runs fine with `On Compress;` in `calcdia.frm`.

Sorry, this is probably not a particularly useful bug report. Can someone try to reproduce this? I have the same behavior on more than one machine. FORM 4.2.0.

EDIT: v4.1-20131025 crashes also.

jodavies:
…le and off compress

In the case if "Off Compress;" the term size was not written into the compress buffer when reading from a file, and term compression is off.

This results in terms which are incomplete or have extra junk on the end going through Generator; the rest is chaos.

Fixes #214

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2016,6 +2016,34 @@ else
 end
 assert result("Zero") =~ expr("0")
 *--#] Issue197 : 
+*--#[ Issue214 :
+#-
+#: MaxTermSize 500
+#: ScratchSize 1K
+#: SortIOSize 1K
+
+Off compress;
+Symbol x,y,z,i;
+CFunction f;
+
+Local test = sum_(i,1,100,f(i*(x+y))*(x+y)^20) - 52824783675150;
+Bracket f;
+.sort
+Keep Brackets;
+
+Identify f(x?) = x;
+.sort
+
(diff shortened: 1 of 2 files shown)


jodavies:
The parform valgrind failure on the new test is present on the master branch also.

jodavies:
@tueda if you are happy with skipping the test under parform valgrind here, I think we can merge this one (after squashing the test skip commit into the other).

tueda:
OK, go ahead.

PR #593, opened 2024-11-28
Fix missed CompressPointer reset in EndSort when without-zlib

### LINKED ISSUE #95: gzip errors
Several times, I have seen errors like

```
1FillInputGZIP: Error in gzip handling of input.
PutIn: We have RetCode = FFFFFFFFFFFFFFFF while reading 124F80 bytes
Program terminating in thread 1 at replace Line 18 -->
```

They are not reproducible, generally I can re-run the calculation and everything goes through without problems. It is hard to know what is to blame here, perhaps they are genuine read errors from the disk (although I have seen this issue on more than one machine).

I don't know if there would be much point, upon receiving an error from zlib, retrying the inflate?

Have you seen this before? Andreas says he does not recall ever seeing a crash like this. Not a very useful report I suppose, but I thought I would mention it in case you have any ideas.

jodavies:
Reset of *AR.CompressPointer was the wrong side of an #ifdef WITHZLIB.

Also fix the --without-zlib build, broken since the split sort allocation patch: 7d0e71db37ff8288fa3e771ec03aa33f8052f7ee

Partially closes #95 ?

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -962,6 +962,40 @@ L F7 = f(10000*g5_);
 #pend_if mpi?
 assert runtime_error?
 *--#] Issue94 : 
+*--#[ Issue95 :
+#-
+#: filepatches 32
+#: largepatches 32
+#: largesize 6250000
+#: maxtermsize 1250
+#: smallsize 1250000
+#: smallextension 2500032
+#: termsinsmall 12500
+
+Off Statistics;
+
+Symbol x,n;
+
+* One fewer term than reported in the Issue, since now the
+* non-x term is in the initial definition of test.
+#define NTERMS "266905"
+
+Local test = sum_(n,1,`NTERMS',x^n) - `NTERMS'*(`NTERMS'+1)/2;
+.sort
+
+* Check all terms present
+Identify x^n?pos_ = n;
+
+Print;
+.end
+# This takes a long time for tform under valgrind.
+# tform -w4 also doesn't crash here anyway (but -w2 does).
+#pend_if valgrind? && threaded?
+# Also doesn't run properly for 32bit form.
+#require wordsize >= 4
+assert succeeded?
+assert result("test") =~ expr("0")
+*--#] Issue95 : 
 *--#[ Issue95b :
 #-
 #:filepatches        4


tueda (approved):
I have confirmed that this PR fixes https://github.com/vermaseren/form/issues/95#issuecomment-353648021 (or 
https://github.com/vermaseren/form/issues/95#issuecomment-2504188080) without zlib. (The source code remains unchanged with zlib.)

Is the suffix `b` in the test case name `Issue95b` intended to suggest a partial fix?

check/fixes.frm
```diff
@@ -962,6 +962,61 @@ L F7 = f(10000*g5_);
...
+#: maxtermsize 1250
+#: smallsize 1250000
+#: smallextension 2500032
+#: termsinsmall 12500
+
+Off Statistics;
+
+Symbol x,n;
+
+* One fewew term than reported in the Issue, since now the
```
tueda, line 979: fewew -> fewer

check/fixes.frm
```diff
@@ -962,6 +962,61 @@ L F7 = f(10000*g5_);
...
+#define NTERMS "266905"
+
+Local test = sum_(n,1,`NTERMS',x^n) - `NTERMS'*(`NTERMS'+1)/2;
+.sort
+
+* Check all terms present
+Identify x^n?pos_ = n;
+
+Print;
+.end
```
tueda, line 990: It seems that this test case requires `wordsize >= 4` because 266905 > 32767.

jodavies:
> Looking back at the discussion, no one actually confirmed that they were using a post - [26793e4](https://github.com/vermaseren/form/commit/26793e439166fe71722f7a59e3c2260801596ee8) binary.

This is not true actually, @alonli1 's log files are FORM 5 beta.

Also of note, is that those two subsequent reports are zerror -3 and not -2 like the original report. This issue should remain open I suppose...

check/fixes.frm
```diff
@@ -962,6 +962,62 @@ L F7 = f(10000*g5_);
...
+# This takes a long time for tform under valgrind.
+# tform -w4 also doesn't crash here anyway (but -w2 does).
+# Also doesn't run properly for 32bit form.
+#pend_if ( valgrind? && threaded? ) || wordsize < 4
```
tueda, line 994: Can you put the wordsize condition to `#require`? `#pend_if` is for a pending status due to FORM bugs to be fixed.

jodavies: Yes, sure.

check/fixes.frm
```diff
@@ -962,6 +962,40 @@ L F7 = f(10000*g5_);
...
+.sort
+
+* Check all terms present
+Identify x^n?pos_ = n;
+
+Print;
+.end
+# This takes a long time for tform under valgrind.
+# tform -w4 also doesn't crash here anyway (but -w2 does).
+#pend_if ( valgrind? && threaded? )
```
tueda, line 993: Sorry to nitpick, but then, you don't need `(` and `)` for the `#pend_if`...

tueda:
I feel the discussions in #95 have become rather lengthy. Maybe we could extract and summarize the remaining problems and create a new issue for them, then this PR could close it.

tueda (approved):
Thanks for the fix. Now it looks good.

jodavies:
Yes, that works for me. I think it is also not impossible that `zerror=-3` there can come from bad hardware, since the issue seems not to be reproducible, and also @jamievicary 's crash was form and not tform, so is a bit more deterministic.

Issue #484, opened 2024-02-29
Infinite loop in TermRenumber


jodavies:
One of our `save` files causes an infinite loop in `TermRenumber`. It seems that `term[0]` and `term[1]` are 0, so `*term` does not match any of the cases, and `t` is set to `term` here:
https://github.com/vermaseren/form/blob/b1f9041b7bb6f5dd4a149c87e70922ad9914b291/sources/store.c#L2502
Then `term` is set back to `t` again here:
https://github.com/vermaseren/form/blob/b1f9041b7bb6f5dd4a149c87e70922ad9914b291/sources/store.c#L2559
and around we go again.

The save file is 100MB, so probably it is not to useful to attach it here...

Presumably the bug is in the saving rather than the loading. Possibly related to #420 ?

PR #594, opened 2024-11-28
Fix rare FindrNumber/TermRenumber crashes when loading saves

### LINKED ISSUE #420: FindrNumber/TermRenumber related crash
Hi everyone,

I have been running many instances of a FORM program acting on files of different dimensions. Most of them run without any issue but a handful keep crashing with the following error:

FindrNumber: n = 1, list has -1 members
Start with 0,0,-1
Renumbering problems
Called from TermRenumber
Program terminating at step2.hel.frm Line 371 -->
 14259.98 sec out of 14313.00 sec

I assumed that this was a RAM allocation error so I tried increasing the setup parameters to 

#: MaxTermSize 20M
#: WorkSpace 3000M

as well as the available RAM (I am working on a cluster and I have to pre-allocate a fixed amount). With 128 GB of available memory the programs are still crashing. 

Also, the programs stop inside a #do i=1,n loop at different values of i, so I would probably exclude a mistake in the code, especially because the vast majority of the times the program works.

Does anyone know what the problem might be or at least what this error is related to? 

Thanks in advance for your help!
 Giulio

jodavies:
Uncommenting this line fixes crashes when loading certain save files, where the sizes of the terms have some interaction with SizeStoreCache.

Changing that value or setting NumStoreCaches to 0 is a works-around also.

Closes #420 and #484

### DIFF
--- a/sources/setfile.c
+++ b/sources/setfile.c
@@ -729,6 +729,8 @@ int AllocSetups(VOID)
 	AM.NumStoreCaches = sp->value;
 	sp = GetSetupPar((UBYTE *)"sizestorecache");
 	AM.SizeStoreCache = sp->value;
+	/* Make sure this is a multiple of sizeof(WORD). */
+	AM.SizeStoreCache = ((AM.SizeStoreCache+sizeof(WORD)-1)/sizeof(WORD))*sizeof(WORD);
 #ifndef WITHPTHREADS
 /*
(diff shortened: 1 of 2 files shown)


tueda (approved):
Good to go ahead. (Because creating simple test cases for this issue seems difficult, adding test cases is not necessary.)

Issue #599, opened 2024-12-08
implicitly declared sets


veretin:
Guten Tag!

Consider the following code

```
c A,B,C;
s j,x,z,y;

g ff = A(2) + B(2) + C(2);

  id A(j?{1,2,3}[x]) = A({ 10, 20, 30}[x]);
  id B(j?{1,2,3}[x]) = B({+10,+20,+30}[x]);
  id C(j?{1,2,3}[x]) = C({-10,-20,-30}[x]);

print;
.end
```

This produces:
'''
   ff =
      A(20) + B(-20) + C(-20);
'''

The term B(-20) is confusing.

Best regards,
Oleg Veretin

tueda:
Thanks for the bug report!

`On names` prints
```
 Sets
   {}: 1 2 3
   {}: 10 20 30
   {}: -10 -20 -30
```
`{+10,+20,+30}` is gone, which may be related to https://github.com/vermaseren/form/issues/554#issuecomment-2461742781.

tueda:
> It seems like the + signs are taken to be -

Yes, this seems a bug in the parser (?)

```form
On names;
S x,j,n;
L F = x^2;
id x^j?{2,}[n] = {+2,}[n];
P;
.end
```
gives `F = -2`, while `id x^j?{+2,}[n] = {2,}[n];` does not match with it.

vermaseren:
The culprit is of course in the routine DoElements in names.c
        while ( *s == '-' || *s == '+' ) { sgn ^= 1; s++; }
That does not look right.

jodavies:
OK, so this one is an easy fix but unfortunately is unrelated to #554 . I didn't get to the bottom of that one yet.

PR #601, opened 2024-12-13
Fix bugs due to empty set in wildcard restriction and + signs

### LINKED ISSUE #554: Segfault by empty implicitly-declared set
This is a bug found in the discussion of #545.

An empty implicitly-declared set results in a segfault.
```form
CF f;
S x;
L F = f(x);
id f(x?{}) = x;
.end
```

<details>
<summary>Valgrind output</summary>

```
FORM 5.0.0-beta.1 (May 21 2024, v5.0.0-beta.1-66-g83e3d41)  Run: Wed Jul  3 20:56:27 2024

==11465== Invalid read of size 4
==11465==    at 0x23B2D2: CheckWild (wildcard.c:2279)
==11465==    by 0x15C4A6: MatchFunction (function.c:1223)
==11465==    by 0x15D8B3: ScanFunctions (function.c:1775)
==11465==    by 0x15967E: FindRest (findpat.c:1133)
==11465==    by 0x1ACE06: TestMatch (pattern.c:346)
==11465==    by 0x1E022D: Generator (proces.c:3964)
==11465==    by 0x1E0A74: Generator (proces.c:4216)
==11465==    by 0x1E2102: Processor (proces.c:406)
==11465==    by 0x150E42: DoExecute (execute.c:858)
==11465==    by 0x17C690: ExecModule (module.c:291)
==11465==    by 0x1D6D57: PreProcessor (pre.c:1041)
==11465==    by 0x21466A: main (startup.c:1690)
==11465==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
```
</details>

jodavies:
This PR fixes #554 and #599 .

The behaviour that I noted at the end of https://github.com/vermaseren/form/issues/554#issuecomment-2461742781 is intended. When creating the {} sets, it finds that they exist already and doesn't create a copy:
https://github.com/vermaseren/form/blob/8abec49f876deae1d13ad159519b301f0b140dc5/sources/names.c#L2471-L2474

Thus the following defines only xs and xs2:
```
#-
off stats;
On names;
CF f;
S x,y;
set xs: ;
set xs2: 1,2;
L F = f(1)+f(2)+f(3);
id f(x?{}) = x;
id f(y?{1,2}) = x^y;
print;
.end
```

There is one point for discussion: the second commit implies that `id f(x?!{}) = x;` also does not match. In principle one could claim that x is indeed not a member of the empty set, so that this should match, but to my knowledge there will never be a case where `x?` does match an empty set (`x?` can not be "nothing" in FORM currently) so it feels a bit odd to match in this case to me.

One could even claim that `x?{}` could be a syntax error and terminate. In that case the `x?!{}` question is irrelevant.

Any thoughts?

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -3313,6 +3313,39 @@ Local test3 = 3;
 #pend_if mpi?
 assert runtime_error?("isnumerical: expression is not yet defined!")
 *--#] Issue577_2 :
+*--#[ Issue599 :
+#-
+On names;
+Off statistics;
+CFunction A1,...,A6;
+Symbol j,x,z,y;
+
+Local ff =
+	+ A1(1) + A2(2) + A3(3)
+		+ A4(1) + A5(2) + A6(3)
+		;
+
+Identify A1(j?{1,2,3}[x]) = A1({ 10, 20, 30}[x]);
+Identify A2(j?{1,2,3}[x]) = A2({+10,+20,+30}[x]);
+Identify A3(j?{1,2,3}[x]) = A3({-10,-20,-30}[x]);
+Identify A4(j?{1,2,3}[x]) = A4({+-10,--20,-+30}[x]);
+Identify A5(j?{1,2,3}[x]) = A5({--10,-+20,+-30}[x]);
+Identify A6(j?{1,2,3}[x]) = A6({-+10,+-20,--30}[x]);
+
+Print;
+.end
+assert succeeded?
+assert result("ff") =~ expr("A1(10) + A2(20) + A3(-30) + A4(-10) + A5(-20) + A6(30)")
+assert stdout =~ exact_pattern(<<'EOF')
+ Sets
+   {}: 1 2 3
+   {}: 10 20 30
+   {}: -10 -20 -30
+   {}: -10 20 -30
+   {}: 10 -20 -30
+   {}: -10 -20 30
+EOF
+*--#] Issue599 : 
 *--#[ PullReq535 :
 * This test requires more than the specified 50K workspace.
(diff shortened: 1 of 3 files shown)


tueda:
In principle, the set for the pattern matching can be changed at compile-time, for example,
```form
CF f,g;
S x;

#$x = f(1,3,5);

#inside $x
  if (match(f(?a$a)));
  endif;
#endinside

L F = f(1,2,3,4,5,6);
L G = g(1,2,3,4,5,6);

repeat id f(?a,x? {`$a',},?b) = x * f(?a,?b);
repeat id g(?a,x?!{`$a',},?b) = x * g(?a,?b);

P;
.end
```
I think it should work even if `$x = f()` such that ``{`$a',}`` is expanded as an empty set.

jodavies:
Do you mean that you would like all of the args of `g` to match `!{}` for `$x=f()`?

Otherwise, this example currently "works as intended" for `$x = f();`.

jodavies:
OK. It is not quite as simple as I thought (just `goto NoMnot;`) because the code at the `NoM` label wants to set `AN.oldvalue = w[3];` and here `w=0`. Probably one needs `AN.oldtype = old2; AN.oldvalue = oldval; goto NoMatch;` as happens also further down in the set restrictions code. It looks like `oldval = w[3];` is set at `TestSet:` for this purpose.

Generally it is not clear in what circumstances various "old" values need to be put back or not.

jodavies:
Rebased, and implements the match for `!{}` with `oldtype` etc put back as I understand it.

tueda (approved):
Looks fine.

Issue #600, opened 2024-12-10
Load balancing problems under valgrind


jodavies:
Some of the (very simple) tests in the suite take a very long time to run with tform under valgrind. The issue is in the load balancing mechanism. On my system, for eg, the following
```
Symbol x;
Local test = (1+x)^40;
.sort
Identify x = 1;
.sort
Print;
.end
```
can take essentially an arbitrarily long time when run with `valgrind tvorm -w4`. (Can someone confirm that they can reproduce this?)

It is hard to debug, since adding print statements in relevant places, gets rid of the problem. But I think I have boiled it down a bit. We are basically at the end of processing. `LoadReadjusted` finds `numtogo = 0` so there are no terms left to move around, but `numbusy > 0`. So it returns 1, and we follow a goto which ultimately just calls `LoadReadjusted` again.

But the problem is the master thread re-acquires the lock on `thr-busy` too quickly, and the workers never get the chance to update their status from BUCKETASSIGNED. So we end up calling `LoadReadjusted` in this loop, millions of times (which you can check with, eg callgrind).

The two commits of https://github.com/jodavies/form/tree/load-balancing each seem to fix the problem, presumably the second is a "more proper" way to do it...

I think the real fix is to change the lock to, eg, a rwlock (which form has macros for already) and then prefer writers when there is contention. But if the above simple fix works and doesn't cause any other issues maybe that is OK.

tueda:
`sleep(0.1)` is interpreted as `sleep(0)` and there is no guarantee that a wait will occur.

`sched_yield` is in [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sched_yield.html), and so probably available on many systems. (Checking its availability in the configure script would be optional but helpful for ensuring portability.)

jodavies:
Interesting, then even the function call is enough to let the other threads through it seems? In principle `sched_yield` doesn't necessarily do anything either, on my system certainly there are cores available to just keep the thread running for `-w4`. In which case `nanosleep` might be more guaranteed to actually do something?

jodavies:
This is resolved by #613

PR #613, opened 2025-01-27
Alleviate lock contention in the load balancing mechanism

jodavies:
Add a small sleep in LoadReadjusted to allow threads to aquire the lock. This enormously speeds up some simple tform tests under valgrind.

See also https://github.com/vermaseren/form/issues/600

This effectively just adds "a bit more waiting" to the already-existing `/* Wait a bit.... */` comment.

### DIFF
--- a/sources/threads.c
+++ b/sources/threads.c
@@ -3309,8 +3309,20 @@ restart:;
 		}
 		goto intercepted;
 	}
+/* This has always been commented. Indeed no lock is held here. */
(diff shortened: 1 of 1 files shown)


tueda (approved):
This looks OK, but is 0.01s really necessary? Though it is a small amount, 10^7 clock cycles might be excessive.

jodavies:
You are probably right. At least on my system sleeping for 1us is also enough to let valgrind run. Let's make it shorter.

PR #622, opened 2025-02-11
ci(refactor): extract build setup logic into setup-build

tueda:
This patch extracts the build setup logic into a composite action called `setup-build` to avoid repeating similar steps in different jobs. The setup process includes installing dependencies and libraries, extracting tarballs, running `autoreconf`, and executing `configure`.

In the future, if we need to elaborate further on setting up libraries (including building and caching), such as MPI or FLINT, we can manage them all in one place.

There are several ways to organize shared logic in GitHub Actions:
- Script files: Hard to track in the GitHub Actions UI because everything appears as a single step when running a script file.
- [Reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows): Can't set environment variables for the caller. (`setup-build` needs to modify `$LIBRARY_PATH` etc.)
- [Composite actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action): Allow environment variable modifications, but [do not show individual steps clearly in the UI](https://github.com/actions/runner/issues/1877). [actionlint](https://github.com/rhysd/actionlint) [doesn't work for composite actions](https://github.com/rhysd/actionlint/issues/46).

This patch adopts the last option. To mitigate the drawback, I added lines like `### Initialize setup ###` in the `run:` field, which will appear in the logs.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -1,0 +1,585 @@
+name: setup-build
+description: >-
+  Set up the build environment by installing necessary tools and dependencies.
+  When necessary, it sets CPATH, LIBRARY_PATH, LD_LIBRARY_PATH, FORMPATH,
+  DISTNAME and CONFIG_ARGS, and runs autoreconf or extracts a tar.gz archive,
+  followed by executing ./configure.
+inputs:
+  features:
+    description: A list of features to enable, separated by spaces. To disable a feature, prefix it with "no".
+    required: false
+    default: ''
+  flint_version:
+    description: FLINT version to build.
+    required: false
+    default: '3.3.1'
+
+runs:
+  using: composite
+  steps:
+    - name: Initialize setup
+      id: setup
+      env:
+        available_features: >-
+          form=false tform=false parform=false vorm=false tvorm=false parvorm=false
+          flint=true ginac=false gmp=true mpfr=true zlib=true zstd=true
+          mpich=false openmpi=false
+          bash=false msys2=false debian=false container=false
+          untar=false autoreconf=false configure=true deploy=false
+          coverage=false hyperfine=false valgrind=false
+          build_flint=false
+          automake=false diffutils=false gcc=true git=false ruby=true wget=false
+          latex=false latex2html=false
+          formlib=false color=false forcer=false series=false
+        # NOTE: Some features and OS image combinations are not implemented.
+      shell: bash
+      run: |
+        ### Initialize setup ###
+
+        # First, parse the input.
+
+        available_features=$(echo "$available_features" | awk '{$1=$1; print}')
+
+        feature_names=''
+        for entry in $available_features; do
+          # Get "name=value".
+          feat_name=$(echo "$entry" | cut -d '=' -f 1)
+          feat_value=$(echo "$entry" | cut -d '=' -f 2)
+          feature_names="$feature_names $feat_name"
+          # Initialize the feature variable.
+          eval "feat_${feat_name}=${feat_value}"
+        done
+        feature_names=$(echo "$feature_names" | awk '{$1=$1; print}')
+
+        input_features=$(echo "${{ inputs.features }}" | awk '{$1=$1; print}')
+        feature_pattern='^('$(echo "$feature_names" | tr ' ' '|')')$'
+        no_feature_pattern='^('$(echo "no$feature_names" | sed 's/ /|no/g')')$'
+
+        for feat in $input_features; do
+          if [[ $feat =~ $feature_pattern ]]; then
+            eval "feat_${feat}=true"
+          elif [[ $feat =~ $no_feature_pattern ]]; then
+            feat=${feat#no}
+            eval "feat_${feat}=false"
+          else
+            echo "Error: unknown feature: $feat" >&2
+            exit 1
+          fi
+        done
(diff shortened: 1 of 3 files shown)


tueda:
Changed from `tr ' ' '|no'` to `sed 's/ /|no/g'` (I misunderstood how `tr` works).

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,357 @@
...
+        # Determine DISTNAME if possible.
+
+        if [ -f ./scripts/git-version-gen.sh ] && command -v git >/dev/null 2>&1; then
+          echo "DISTNAME=form-$(./scripts/git-version-gen.sh -r | sed '2q;d' | sed 's/^v//')" >>"$GITHUB_ENV"
+        fi
+
+    # awalsh128/cache-apt-pkgs-action does not allow empty packages argument.
+    # See: https://github.com/awalsh128/cache-apt-pkgs-action/issues/149
+    # As a workaround, we need to check if any packages need to be installed.
+    - name: Install dependencies (Ubuntu)
```
tueda, line 199: This can be simplified if https://github.com/awalsh128/cache-apt-pkgs-action/pull/154 is merged.

tueda:
Updated.

jodavies:
This looks to be building and caching flint also on Ubuntu, right? Then #703 will go in easily?

The actions setup looks much cleaner with this!

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,506 @@
...
+      if: runner.os == 'Windows' && runner.arch == 'X64' && steps.setup.outputs.msys2 == 'true'
+      uses: msys2/setup-msys2@v2
+      with:
+        update: true
+        msystem: mingw64
+        install: >-
+          make
+          mingw-w64-x86_64-gcc
+          mingw-w64-x86_64-ruby
+          ${{ steps.setup.outputs.gmp == 'true' && 'mingw-w64-x86_64-gmp' || '' }}
```
tueda, line 293: I removed `diffutils` from the code in #640 because I couldn't recall why I had added it, but it seems that the FLINT configure script still relies on it:
```
2025-08-26T05:13:26.3544758Z ./configure: line 6794: cmp: command not found
2025-08-26T05:13:26.3747836Z ./configure: line 6794: cmp: command not found
2025-08-26T05:13:27.9965375Z ./configure: line 8551: diff: command not found
```
I'm not sure whether these errors are actually critical, though.

jodavies: `cmp` and `diff` are used in various places in the generated configure script, but at first glance I am not really sure what the code is really doing. It should be no harm to put diffutils back in, anyway?

tueda: Right. Adding `diffutils` should be fine.

The relevant code in the FLINT configure script comes from `libtool.m4`. Without `diff`, I think `-fno-rtti -fno-exceptions` are not set in this case.

.github/workflows/deploy.yml
```diff
@@ -351,14 +217,14 @@ jobs:
               esac
             done
           )
-          cp -r doc/manual/manual $distname
-          tar -c $distname/* | gzip -c -9 > $distname.tar.gz
+          cp -r doc/manual/manual ${{ env.DISTNAME }}
+          tar -c ${{ env.DISTNAME }}/* | gzip -c -9 > ${{ env.DISTNAME }}.tar.gz
```
tueda, line 221: The file name is incorrect and conflicts with the tarball.

jodavies:
Right, this would speed things up (a bit) if the binaries are cached. At some point it might be worth taking a look at the slowest of the tests under valgrind, and trying to make an easier test which hits the same bug, if we would like the CI to run faster (since in the future, we would like to add a lot more tests for better coverage).

tueda:
I just meant removing the five `run: make -C sources -j 4 ${{ matrix.bin }}` lines :-)
Caching the binaries would be tricky, but potentially useful.

jodavies:
Then I misunderstood. In the context of your changes here it does make sense to remove as much duplicate build code as you can from the various test jobs indeed.

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,511 @@
...
+
+        mkdir -p build/flint
+        cd build/flint
+        # We use the tarball from GitHub Releases.
+        # See: https://github.com/flintlib/flint/issues/2311
+        wget https://github.com/flintlib/flint/releases/download/v${{ inputs.flint_version }}/flint-${{ inputs.flint_version }}.tar.gz
+        tar -xf *.tar.gz
+        rm *.tar.gz
+        cd flint-*
+        ./configure --prefix=$GITHUB_WORKSPACE/lib/flint --enable-static
```
tueda, line 416: It appear that the FLINT configure script automatically selects the CPU architecture to optimise performance, which results in non-portable binaries. To address this problem, the `--host` option (for `>= 3.2.0`) must be specified with the correct host triplet  (`$(gcc -dumpmachine)` may be enough). Homebrew uses the [this code](https://github.com/Homebrew/homebrew-core/blob/7d54986bb799e750bc2fcd61fb0977e929e9ae12/Formula/f/flint.rb#L51-L56).

PR #644, opened 2025-06-02
Flint interface v1

jodavies:
The large commit is a squash of most of the individual commits here https://github.com/jodavies/form/tree/flint-prf (where the history can be retained) to keep things neat on the master branch. The first 3 fix some bugs in poly's `inverse_` (but not everything, see below) and corrects the return value of `EndSort` in some cases. The last test adds some tests which are not too painful for valgrind runs on github actions. Flint is enabled by default, so it also passes the existing test suite.

The flint interface currently implements the following, with much improved performance over poly, particularly for multivariate cases.
`PolyRatFun`
`FactArg`
`FactDollar`
`div_`
`rem_`
`mul_`
`gcd_`
`inverse_`

It does not yet support expression factorization or modulus mode; I plan to add those in later, but I would like to merge this before the workshop next week so that there is more time for testing. If someone can have a bit of a look through and test a bit beforehand, that would be great.

Caveats:
 - I still have some tests cases where poly's `inverse_` crashes, so something is still wrong there (but the flint version seems OK). But this is a separate issue to some extent. For this reason `inverse_` is not yet part of the extra tests.
 - A test case is disabled under valgrind, because it has errors due to flint for versions < 3.2.1 (see https://github.com/flintlib/flint/issues/1833#issuecomment-2927746256). There are no reported errors or test failures on the flint side for this, but it is UB.
 - I have not checked this is valgrind clean under parform. Also the extra tests are disabled for parform due to timeouts (I think parform doesn't like the large number of Locals and dollar vars the tests create).
 - Do we want to somehow run the existing test suite using both flint and poly?


jodavies:
It is not so nice that FORM built without flint can not run the new tests: for that I could use for eg `assert succeeded? || warning?("FORM was not built with FLINT support.")`, which would then cause the new tests to "compare poly against itself".

The final bullet point above could be resolved by duplicating every test which uses poly, and making separate poly and flint versions (which have `Off flint;` and `On flint;` added). This should produce faster CI times than simply running the entire test suite twice, in each mode.

.github/workflows/test.yml
```diff
@@ -38,7 +41,7 @@ jobs:
       - name: Install dependencies
         uses: awalsh128/cache-apt-pkgs-action@v1
         with:
-          packages: libmpfr-dev libzstd-dev
+          packages: libmpfr-dev libzstd-dev libflint-dev
```
tueda, line 44: Minor suggestion: The packages could be sorted alphabetically.

sources/Makefile.am
```diff
@@ -84,6 +84,12 @@ SRCBASE += \
 	fwin.h
 endif
 
+if WITHFLINT
+SRCBASE += \
+	flintinterface.cc \
+	flintwrap.cc
```
tueda, line 90: We need `flintinterface.h` (in tarball distributions).

PR #649, opened 2025-06-13
Warn by default if parallel execution is vetoed due to dollar variables

a-maier:
Before this commit FORM only warns when disabling parallel execution in a module if `on allwarnings` is set. This makes sense, since for most of the causes it is not straightforward to modify the code in a way that re-enables parallel execution. However, modifying a dollar variable without a corresponding `moduleoption` statement is often a trivial bug and the user should be warned by default.

### DIFF
--- a/check/examples.frm
+++ b/check/examples.frm
@@ -519,6 +519,7 @@ EOF
     #$a = 0;
     if ( count(x,1) > $a ) $a = count_(x,1);
     Print "      >> After %t the maximum power of x is %$",$a;
+    ModuleOption noparallel; * suppresses noparallel warning
     #write "     ># $a = `$a'"
     .sort
     #write "     ># $a = `$a'"
@@ -729,6 +730,7 @@ assert result("G") =~ expr("
     id,all,$t*replace_(<p1,p1?>,...,<p5,p5?>) =
          $t*replace(<p1,q1>,...,<p5,q5>);
     Print +s;
+    ModuleOption local $t; * allow parallel execution
     .end
     assert succeeded?
     assert result("F") =~ expr("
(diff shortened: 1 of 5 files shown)


tueda:
I think adding `ModuleOption` statements for these failed tests is a reasonable approach, unless you expect all of them to test the new warning behaviour.

For tests in `examples.frm`, it would be helpful to include comments to clarify why the `ModuleOption` statement is added to the corresponding example in the reference manual, like:
```form
ModuleOption noparallel; * suppresses noparallel warning
```

jodavies:
For the tests where you added `local` rather than `noparallel` it doesn't look like it meaningfully affects what the tests are checking, so that should be OK. I can't reproduce the test failure for `PolyandFact_14` locally?

Bit of a nitpick: could you remove the empty lines between the folds, so that the new tests match the rest? The modifed "examples" tests could also have their code modified in the manual.

a-maier:
> I can't reproduce the test failure for PolyandFact_14 locally?

Sorry, I guess that one shouldn't run in parallel mode because it can change the order of the output lines.

> For the tests where you added local rather than noparallel it doesn't look like it meaningfully affects what the tests are checking, so that should be OK.

Right. I also thought it might be useful to check that they still work properly with parallel execution.

 >The modifed "examples" tests could also have their code modified in the manual.

I have updated two out of three, but I am a bit unsure about the  `DolVars_1`. The section explaining dollar variables and  `moduleoption` for TFORM and PARFORM only comes after it in the manual.

jodavies:
> Sorry, I guess that one shouldn't run in parallel mode because it can change the order of the output lines.

Right, now I see it: the terms are of course processed in an arbitrary order by tform.

jodavies:
If you rebase on master and squash the commits a bit (adjust existing tests, add the warnings, add tests of the warnings) I think this can go in, if @tueda is happy with the test modifications?

PR #652, opened 2025-06-14
fix: memory leak in InsideDollar

jodavies:
Fixes the leak discovered in #648


jodavies:
Updated: I added a test, this leak is specifically when you work inside a dollar variable that is zero.

PR #653, opened 2025-06-15
Mark Terminate() and a few more functions as noreturn

magv:
This helps the compiler identify which code paths are taken rarely, and move them out of the way from hot paths. I'd expect a marginal (below 1%) performance improvement, if there was a benchmark suite. The `NORETURN` macro currently expands into `__attribute__((noreturn))`, which GCC and clang understand, or nothing, if the configure script figured out that the compiler doesn't know the syntax. In principle, C11 and C++11 both support a noreturn attribute (with different syntax), but I wasn't sure what is the minimum language version requirement.

There's the issue of the Crash() function that crashes or not depending on the flags. I did not mark it as noreturn. I think that one deserves a separate consideration (i.e. why is it used in `sort.c`?).


jodavies:
@magv can you fix the conflicts since the merge of #674 ?

PR #654, opened 2025-06-15
Include FLINT in release binaries

tueda:
The binaries are now built with the `--with-flint` configure option and statically linked with FLINT.


tueda:
Hopefully, the last commit resolves the `WORD` macro conflict on Windows.

PR #674, opened 2025-06-19
Use more stdint.h types, and drop ILP32 support

magv:
Use `stdint.h` instead of INT{16,32,64}. Make WORD always int32_t, and drop ILP32 support. As discussed in #670.


sources/minos.c
```diff
@@ -295,13 +295,11 @@ int ReadIndex(DBASE *d)
 	Allocate the pieces one by one (makes it easier to give it back)
 */
 	if ( d->info.numberofindexblocks <= 0 ) return(0);
-#ifndef WORDSIZE32
```
jodavies, line 298: This one is an `ifndef`, but it doesn't harm to anyway make this check.

magv: Yeah, I thought it's a sanity check that makes sense in 64-bit mode too.

sources/declare.h
```diff
@@ -384,20 +384,20 @@ TP=T+1;while(TP<TT){if(*TP==AR.PolyFun){TP[2]|=(DIRTYFLAG|MUSTCLEANPRF);}TP+=TP[
  */
 static inline unsigned int IntAbs(int x)
 {
-	if ( x >= 0 ) return x;
-	return(-((unsigned int)x));
+	if ( x >= 0 ) return (unsigned int)x;
+	return((unsigned int)(-x));
```
jodavies, line 388: Doesn't this break the function? Here it should correctly handle INT_MIN, hence the cast followed by unary -.

magv: You're right. I'll undo this. I though that casting a negative int to unsigned is undefined, but apparently it is.

magv:
OK, so this is the current proposed set of changes to use stdint.h more. This is what was discussed in @670, except I've decided not to remove ILP32 support for not, because it turned out to be very helpful in catching WORD vs int distinction in the code (something I'd wish the compiler would do for us, but it wouldn't). Instead I've slightly simplified the detection of 32-bit platforms using stdint.h alone, and updated all the code that did `#ifdef ILP32` to `#if BITSINWORD == 16`. I would, of course, still support deprecating the 32-bit support, but keeping the (minimal set of) tests for it running would still be beneficial.

sources/minos.h
```diff
@@ -59,19 +59,22 @@
...
 */
 
-#define MAXBASES 16 
-#ifdef WORDSIZE32
-#define NUMOBJECTS 1024
-#define MAXINDEXSIZE 1000000000L
-#define NAMETABLESIZE 1008
-#define ELEMENTSIZE 256
+#if BITSINWORD == 32
+	#define MAXBASES 16
```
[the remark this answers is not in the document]
magv: Oh, I messed this one up. I'll fix it... by removing MAXBASES.

tueda:
If I'm not mistaken, `off_t` is 32 bits in 64-bit MSVC, but it may be OK to remove the large file support warning (anyway there would be more issues for LLP64 and no one cares).

PR #676, opened 2025-06-20
Fix float printing

cbmarini:
This fixes the float printing issue

```
#StartFloat 50
Local F = 1/3;
ToFloat;
Print;
.end
```
that resulted in 
```
   F =
      3.3e-01;
```
The correct result is now
```
   F =
      3.333333333333333e-01;
```
Furthermore set the default MZV weight `MAXWEIGHT` to 0 instead of 40.

### DIFF
--- a/check/features.frm
+++ b/check/features.frm
@@ -617,15 +617,15 @@ Print;
 .end
 #pend_if wordsize == 2
 assert result("PI0") =~ expr("1.0e+00*a*b")
-assert result("PI1") =~ expr("3.14159265358979323846e+00")
-assert result("PI2") =~ expr("9.42477796076937971538e+00")
+assert result("PI1") =~ expr("3.1415926535897932385e+00")
+assert result("PI2") =~ expr("9.4247779607693797154e+00")
 assert result("PI3") =~ expr("5.4413980927026535518e+00")
-assert result("PI4") =~ expr("3.10062766802998201755e+01")
-assert result("PI5") =~ expr("3.14159265358979323846e+00*a*b")
-assert result("PI6") =~ expr("1.77245385090551602731e+00")
-assert result("EE1") =~ expr("2.71828182845904523537e+00")
-assert result("EE2") =~ expr("8.53973422267356706549e+00")
-assert result("EM1") =~ expr("5.77215664901532860607e-01")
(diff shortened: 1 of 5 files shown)


cbmarini:
I see that the tests that printed some of the constants like `pi_` now also need to be updated.

cbmarini:
I changed `SetupMPFTables` such that it uses `GMP`'s `mpf_init`. At the moment, the MZV tables are set with variable precision (see the comment at the beginning of `SetupMZVTables`). Do we still want this, or is it better to set the tables also with the highest precision using `mpf_init`?

check/features.frm
```diff
@@ -617,15 +617,15 @@ Print;
...
-assert result("PI6") =~ expr("1.77245385090551602731e+00")
-assert result("EE1") =~ expr("2.71828182845904523537e+00")
-assert result("EE2") =~ expr("8.53973422267356706549e+00")
-assert result("EM1") =~ expr("5.77215664901532860607e-01")
+assert result("PI4") =~ expr("3.1006276680299820175e+01")
+assert result("PI5") =~ expr("3.1415926535897932385e+00*a*b")
+assert result("PI6") =~ expr("1.7724538509055160273e+00")
+assert result("EE1") =~ expr("2.7182818284590452354e+00")
+assert result("EE2") =~ expr("8.5397342226735670654e+00")
+assert result("EM1") =~ expr("5.772156649015328606e-01")
```
jodavies, line 628: Does it make sense that there is a digit less here?

cbmarini: If you look at the next digits, one has `5.77215664901532860606e-01`. So it looks like it got truncated: `60606` becomes `6060` by `gmp_snprintf` in PrintFloat. Then the parsing of FORM drops the last 0. So in total, one gets one digit less compared to the other examples in the test.

cbmarini:
For the new test that evaluates all MZVs up to weight 6, we could consider increasing the weight. On my laptop, evaluating up to weight 10 only takes about 0.25 seconds.

cbmarini:
@vermaseren, @tueda, @jodavies do you think this can be merged? This will also help when making more tests for the floating point system, see e.g. #492.

PR #677, opened 2025-06-20
ci: make valgrind and coverage require successful check to start

jodavies:
Avoid burning CPU time and clogging up the runners if the normal checks fail.

Also separate parform from check: it takes a long time


tueda:
Do we need to make the `valgrind-check` and `coverage` jobs require the successful completion of the `check-i386` job?

PR #678, opened 2025-06-22
feature: add "On humanstats;" for human-readable units in stats

jodavies:
Off by default.


tueda:
Can you add tests to cover (at least some of) the new code in https://coveralls.io/builds/74539873/source?filename=sources%2Fsort.c?

jodavies:
Ah, sorry, I did intent to add a test case. I'll add it separately.

Issue #675, opened 2025-06-20
transform dropargs -- whole argument list is dropped when having fewer arguments


adamkardos:
Hello,

I bumped into this: whenever I want to use transform with dropargs and it is applied on a function
with fewer arguments than the indicated range to drop the whole argument list disappears.

MWE:

```
cfunction f, g, h;
symbols a,b,c;

local F = f(a) + g(a, b) + h(a, b, c);

transform,f,dropargs(2,last);
transform,g,dropargs(2,last);
transform,h,dropargs(2,last);

print +ss;
.end
```

this produces:
```
  F =
       +
         f
       +
         g(a)
       +
         h(a)
      ;
```

I am using: 207386593d90a83e3760e1db60d827151479e1e2

Thanks so much!

jodavies:
This one is similar to #267 , I’ll take a look

jodavies:
What happens, is that `(2,last)` produces `(2,1)` which is then ordered to `(1,2)`. Then later it checks if the second number is greater than the total number of arguments, and clamps it.

Here we should "do nothing" if either number is greater than the total of arguments.

jodavies:
Without modifying `FindRange` it isn't possible to know if a number came from `last` or was directly supplied by the user. The manual is not specific regarding what happens in these cases, and not all operations are consistent (#267) so maybe the questions are for @vermaseren :
- what should `transform f dropargs(1,2);` do to `f(1)` ?
- what should `transform f dropargs(2,last);` do to `f(1)` ?

And the same questions apply to addargs, mulargs too.

Issue #267, opened 2018-02-16
Crash in Transform f addargs(1,last) if f appears with no arguments


jodavies:
Hello,

This is not really a bug as such, but something I found and thought I should report. I would expect the following code to simply "do nothing",
```
#-
CFunction f;
Local test = f;
Transform f addargs(1,last);
Print +s;
.end
```
but rather it gives the misleading error
```
Illegal $ value in range while executing transform statement.
```

Of course one can avoid it with
```
Identify f = 0;
```
or similar, before the transform statement.

Thanks,
Josh.

jodavies:
Now I hit this again with mulargs. Let's make this give a sensible error message at least, but I would prefer it to just "do nothing" in the 0-arg cases. Any thoughts?

jodavies:
For this I will prepare a PR which "does nothing" if transform encounters a function with no args and "last" has been used. Experimenting more, this is consistent with the behaviour if a fixed number of args is provided, for eg `Transform f addargs(1,2);` just ignores `f` (and `f(1)`). Also I will give a better error message if the user passes a range spec that is `<= 0` directly (currently it also complains about dollar vars).

While looking at the functions in transform I also saw that there is a `LOCK` placed and not released in `RunPermute` if a dollar var is used for the cycles and a moduleoption is present that is not "local". The following leads to a deadlock, for example (honestly I am not sure when such dollar variable use is what you would want...)
```
CFunction f;
Symbol a1,...,a7;
Symbol x,n;

Local test = f(a1,a2,a3,a4,a5,a6,a7);
.sort

$doln = 1;
Transform f permute($doln,3,5)(2,6);

ModuleOption maximum $doln;
Print;
.end
```

jodavies:
The commits here fix up the main part of the issue: https://github.com/jodavies/form/tree/transform-noargs

I actually don't like the discrepancies in how addargs and mulargs treat their ranges. Compare:
https://github.com/form-dev/form/blob/207386593d90a83e3760e1db60d827151479e1e2/sources/transform.c#L2717-L2720 to
https://github.com/form-dev/form/blob/207386593d90a83e3760e1db60d827151479e1e2/sources/transform.c#L2795-L2800

Specifically, addargs ignores functions which have fewer arguments than the upper range limit, but mulargs treats it like "last". This should probably be considered a bug, but fixing it may well break user code.

Any thoughts?

PR #688, opened 2025-07-04
Fix crash in transform addargs/mulargs if function has no args

jodavies:
Just so I don't forget, here is the fix for #267 but it is waiting for discussion on what the behaviour should be regarding functions with fewer arguments than are specified, see also #675 .


jodavies:
I'll merge this fix on its own, and in the future we can think about the behaviour discussed in #675 , potentially changing it and also making the behaviour of addargs/mulargs/dropargs... more uniform.

PR #689, opened 2025-07-04
fix: repeat if sum creates dummy indices

### LINKED ISSUE #615: Unexpected behaviour of repeat?
Hello,

I discovered some (for me) unexpected behaviour of the repeat environment. The following code replaces only one index pair (I'd have expected it to replace both):

```
cf Gam;
i nu4,nu6,mu1;
i fli1,fli2;

l exp = Gam(1,nu4,nu6)*Gam(2,nu4,nu6) ;

repeat;
  if (match(Gam(fli1?,?x,mu1?!dummyindices_$sumind,?y)*Gam(fli2?,?z,mu1?,?w))) sum $sumind;
endrepeat;

print;
.end
```
gives
```
   exp = Gam(1,nu4,N1_?)*Gam(2,nu4,N1_?);
```
while only when including the statement twice, like
```
cf Gam;
i nu4,nu6,mu1;
i fli1,fli2;

l exp = Gam(1,nu4,nu6)*Gam(2,nu4,nu6) ;

repeat;
  if (match(Gam(fli1?,?x,mu1?!dummyindices_$sumind,?y)*Gam(fli2?,?z,mu1?,?w))) sum $sumind;
  if (match(Gam(fli1?,?x,mu1?!dummyindices_$sumind,?y)*Gam(fli2?,?z,mu1?,?w))) sum $sumind;
endrepeat;

print;
.end
```
it gives
```
   exp = Gam(1,N1_?,N2_?)*Gam(2,N1_?,N2_?);
```
Is there a reason for that?

Thanks!

jodavies:
This fixes #615 , but not the other "mode" of `sum` (`TYPESUMFIX`), i.e. if we have something like `sum $sumind,1,2,3,4;`.

I am not sure yet how to fix things in that case: it is inside the call of Generator here
https://github.com/form-dev/form/blob/d9a3230f13d5f9199653d07584067673ce239de7/sources/proces.c#L3529
where we need to repeat the statements of the repeat block, right? The term we started with disappears. Setting `RepCount` before or after the call of `Generator` here doesn't work.

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -3794,6 +3794,32 @@ assert stdout =~ exact_pattern(<<'EOF')
    {}: -10 -20 30
 EOF
 *--#] Issue599 : 
+*--#[ Issue615 :
+#-
+Off stats;
+
+Index i1,i2,i3,i4;
+CFunction f;
+Set sumind: i2,i3;
+
+Local test1 = f(i1,i2,i3,i4)^2;
+Local test2 = f(i1,i2,i3,i4)^2;
+
+Repeat;
+	If (Match(f(?a,i1?sumind$sum,?b))) Sum $sum;
+EndRepeat;
+InExpression test2;
+	Repeat;
+		If (Match(f(?a,i1?!dummyindices_$sum,?b))) Sum $sum;
+	EndRepeat;
+EndInExpression;
+
+Print;
+.end
+assert succeeded?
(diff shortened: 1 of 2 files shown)


jodavies:
Updated to add a test. I'll merge this, with the caveat that the `sum $sumind,1,2,3,4;` mode of `sum` still doesn't work. Fixing this can be investigated separately. There are anyway ways to "workaround" this in a FORM script.

PR #692, opened 2025-07-09
test: simple test of the "On Humanstats;" output

jodavies:
Also tweak the threshold for the display of "<1 K" etc. We should never see "0 K" printed.

### DIFF
--- a/check/features.frm
+++ b/check/features.frm
@@ -790,6 +790,27 @@ assert result("MZV6") =~ expr("
        + 1.017343061984449139715e+00*mzv(6)
 ")
 *--#] evaluate_all_mzv_2-6 :
+*--#[ humanstats :
+#-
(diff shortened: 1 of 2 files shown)


jodavies:
It is floating because initially, I had it printing "0.1 K" etc, before changing that. It could also just be an integer type (and print "<1 K" for numbers up to 999, which would also be OK for me if you prefer.

Or it can stay as a float (and I change the type of the literal). (s)(n)printf don't have a specifier for float vs double, as far as I know, the conversion there can be avoided by just changing to double everywhere.

tueda:
OK. Maybe we can stay with `float`, using `1000.0f` and `0.5f`.
(`float` passed to `snprintf` is automatically converted to `double` by the promotion rule).

PR #701, opened 2025-08-19
perf: tform: remove default-enabled compare count

jodavies:
For scripts which are dominated by level-0 sorting, counting the compares (and not even printing the result) leads to a large performance impact.

Disabled the counting by default.

The per-thread counter still leads to a performance impact, though it is smaller.

---------

This counter was introduced in f1b83ae, which adds the partial binary search to SplitMerge. There a 1.5% performance improvement is claimed for mincer. Possibly this was determined with a single-threaded test?

This change (default: no counter) compared to current master:

| Test | master (s) | Speedup |
|------|--------------|---------|
|trace|5.09|4.51x|
|mincer|70.57|1.28x|
|mass-fact|59.38|1.12x
|forcer|122.23|1.02x|
|mbox1l|2.73|1.00x|

`forcer` and `mbox1l` tests spend a lot of time sorting polyratfun, so they have much less level-0 sorting time.

Please try your own performance tests and let me know if this is observed on other systems. Here I use 24 threads.

### DIFF
--- a/sources/sort.c
+++ b/sources/sort.c
@@ -72,7 +72,13 @@ extern LONG nummallocs;
 extern LONG numfrees;
 #endif
 
-LONG numcompares;
+//#define COUNTCOMPARES
+#ifdef COUNTCOMPARES
+	// This needs to be large enough for the number of threads.
+	// It is hardcoded here, but 1024 should be enough.
+	// Enabling this has a performance impact.
+	LONG numcompares[1024];
+#endif
 
 /*
   	#] Includes : 
@@ -659,7 +665,13 @@ WORD NewSort(PHEAD0)
 	}
 	if ( AR.sLevel == 0 ) {
 
-		numcompares = 0;
+#ifdef COUNTCOMPARES
+#ifdef WITHPTHREADS
+		numcompares[AT.identity] = 0;
+#else
+		numcompares[0] = 0;
+#endif
+#endif
 
 		AN.FunSorts[0] = AT.S0;
 		if ( AR.PolyFun == 0 ) { AT.S0->PolyFlag = 0; }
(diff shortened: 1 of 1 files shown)


sources/sort.c
```diff
@@ -72,7 +72,13 @@ extern LONG nummallocs;
 extern LONG numfrees;
 #endif
 
-LONG numcompares;
+//#define COUNTCOMPARES
+#ifdef COUNTCOMPARES
+	// This needs to be large enough for the number of threads.
+	// It is hardcoded here, but this just is for debugging anyway.
+	// Enabling this has a performance impact.
+	LONG numcompares[64];
```
magv, line 80: I know quite a few machines with more than 64 threads :)

jodavies: Certainly, I wasn't sure of a sensible "default". Since you have to enable this at compile time (and basically only when you want to debug / develop sorting algorithms) it is basically up to the person to set correctly for their purposes.

magv: Can't this be dynamically allocated? Isn't there a per-thread structure somewhere already where this can be slotted in?

jodavies: Yes, it could live in AT in principle.

magv:
That's a crazy improvement for such a small change. Well done finding it.

Why is `form` 10% slower though?

jodavies:
Ah, looks like fu.frm is basically what I've called "trace" (it comes from aap1000 also). Your slowdown for form is unexpected, though? I don't observe this.

jodavies:
I modified the per-thread counters to 1024. I don't want to add to AT with an ifdef because the nested padding macros are a pain, and this is not something that should normally be enabled.

jodavies:
If I re-run the forcer test with `polyratfunexpand=10` then there is a 10% improvement there also. I think we can merge this in the current state.

We should think about a standard set of performance regression tests which are not too time consuming, probably.

PR #720, opened 2025-10-01
fix: memory leak in pre-processor dollar var increment

### LINKED ISSUE #666: Pre-processor dollar increment leaks memory
This is extracted from the "minceex" test that I wanted to add to `check/extra`, which is not valgrind clean.
```
#$repcount = 1;

Local test = 1;

#do i = 1,10
   Multiply 2;
   .sort:iter `$repcount++';
#enddo

print;
.end
```
produces:
```
32 bytes in 1 blocks are possibly lost in loss record 47 of 168
   at 0x4846828: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x26FEDF: Malloc1 (tools.c:2322)
   by 0x17C2EA: WriteDollarToBuffer (dollar.c:606)
   by 0x271A2B: OpenStream (tools.c:418)
   by 0x21503E: GetChar (pre.c:288)
   by 0x2159C8: LoadInstruction (pre.c:1279)
   by 0x1B49A3: ModuleInstruction (module.c:82)
   by 0x216EC9: PreProcessor (pre.c:1021)
   by 0x25761A: main (startup.c:1745)

288 bytes in 9 blocks are definitely lost in loss record 89 of 168
   at 0x4846828: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x26FEDF: Malloc1 (tools.c:2322)
   by 0x17C2EA: WriteDollarToBuffer (dollar.c:606)
   by 0x271A2B: OpenStream (tools.c:418)
   by 0x21503E: GetChar (pre.c:288)
   by 0x2159C8: LoadInstruction (pre.c:1279)
   by 0x1B49A3: ModuleInstruction (module.c:82)
   by 0x216EC9: PreProcessor (pre.c:1021)
   by 0x25761A: main (startup.c:1745)
```
Here `OpenStream` allocates memory which is then put into a STREAM struct. My understanding is that `GetInput` from this stream should eventually reach the end and call `CloseStream` automatically, but the presence of the `++` causes problems.

jodavies:
This memory is allocated in WriteDollarToBuffer as "DollarOutBuffer" and then stored in stream->buffer by OpenStream. We must free it in CloseStream in the PRERAISEAFTER/PRELOWERAFTER case.

This closes #666 .

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -3943,6 +3943,19 @@ Evaluate 1;
 #pend_if wordsize == 2
 assert compile_error?("should be a built in function that can be evaluated numerically.")
 *--#] Issue664 :
+*--#[ Issue666 :
+#-
+#$repcount = 1;
+Local test = 1;
+Multiply 2;
+.sort:iter `$repcount++';
+Multiply 2;
+.sort:iter `$repcount++';
+Print;
+.end
+assert succeeded?
+assert result("test") =~ expr("4")
+*--#] Issue666 :
 *--#[ Issue668_1 :
 * Check error message for invalid setup parameter:
 #:x
--- a/sources/tools.c
+++ b/sources/tools.c
@@ -736,6 +736,8 @@ STREAM *CloseStream(STREAM *stream)
 		if ( stream->afterwards == PRERAISEAFTER ) x = 1;
 		else x = -1;
 		DollarRaiseLow(stream->pname,x);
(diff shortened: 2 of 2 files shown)


jodavies:
Once this is in, I can add "minceex" based tests to check/extra.

PR #726, opened 2025-10-07
test: add tests based on mincer2 and minceex

jodavies:


jodavies:
@tueda are you happy with the changes in `actions.yml`? It means we `wget https://github.com/form-dev/form-packages/archive/refs/tags/v1.0.0.tar.gz` twice, but allows one to enable color and mincer independently in principle (not that we currently need to do this).

tueda:
I think this is a reasonable choice for now. In principle, they could be combined, but doing so requires some changes and a decision on whether to use a single package or write more complex code that enables a single download to be reused for the two packages.

PR #729, opened 2025-10-09
flint: perf: use mpoly for sparse univariate computations

jodavies:
This improves performance for things like "gcd_(1-x^20,1-x^1000000)" since FLINT poly has a dense representation whereas mpoly is sparse. "Real life" benchmarks such as forcer or minceex are unchanged, they have dense polynomials.

----

Any thoughts on this? I have determined the threshold "experimentally" by benchmarking `gcd_(1-x^20,1-x^N)` for a range of `N`. Of course different computations probably have a different threshold... this kind of heuristic usually has tricky cases.

### DIFF
--- a/sources/flintinterface.cc
+++ b/sources/flintinterface.cc
@@ -38,6 +38,13 @@ static_assert(sizeof(slong) == sizeof(int64_t), "flint interface expects slong i
 */
 
 /*
+ * FLINT's univariate poly has a dense representation. For sufficiently sparse polynomials it is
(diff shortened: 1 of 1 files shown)


jodavies:
Updated, the term counting was not correct in fact.

I ran a few benchmarks also for `mul_` and `div_`, the optimal threshold for those is not exactly the same (for the tested polynomials) as for `gcd_` but the value I chose before is broadly OK.

Issue #251, opened 2017-12-11
Crash in bracket substitutions in RHS with denominators


tueda:
```
S x,a0,...,a3;
L F = <a0*x^0>+...+<a3*x^3>;
B x;
.sort
L G =
  + 1 / F[1]
  + x * (-F[x]) / F[1]^2
  + x^2 * (F[x]^2 - F[1]*F[x^2]) / F[1]^3
  + x^3 * (-F[x]^3 + 2*F[1]*F[x]*F[x^2] - F[1]^2*F[x^3]) / F[1]^4
;
.end
```
```
FORM 4.2.0 (Nov 30 2017, v4.2.0-24-g1514492) 64-bits  Run: Mon Dec 11 15:45:59 2017

...

Program terminating at test.frm Line 5 --> 
```

Valgrind output:
```
...

==32089== Invalid write of size 1
==32089==    at 0x4FCC12: simp3btoken (token.c:1707)
==32089==    by 0x4FE6AC: tokenize (token.c:565)
==32089==    by 0x42A8E4: CompileAlgebra (compiler.c:506)
==32089==    by 0x413C5D: DoExpr (comexpr.c:252)
==32089==    by 0x413F7E: CoLocal (comexpr.c:66)
==32089==    by 0x4261FC: CompileStatement (compiler.c:634)
==32089==    by 0x4B02FE: PreProcessor (pre.c:1045)
==32089==    by 0x4E8ECF: main (startup.c:1605)
==32089==  Address 0x617c450 is 0 bytes after a block of size 400 alloc'd
==32089==    at 0x4C2AB96: malloc (vg_replace_malloc.c:299)
==32089==    by 0x4FF9B8: Malloc1 (tools.c:2230)
==32089==    by 0x501880: DoubleBuffer (tools.c:2887)
==32089==    by 0x4FE649: tokenize (token.c:558)
==32089==    by 0x42A8E4: CompileAlgebra (compiler.c:506)
==32089==    by 0x413C5D: DoExpr (comexpr.c:252)
==32089==    by 0x413F7E: CoLocal (comexpr.c:66)
==32089==    by 0x4261FC: CompileStatement (compiler.c:634)
==32089==    by 0x4B02FE: PreProcessor (pre.c:1045)
==32089==    by 0x4E8ECF: main (startup.c:1605)
==32089==

... and other errors in simp3btoken
```

PR #731, opened 2025-10-14
fix: use a scratch buffer in simp3btoken

jodavies:
Issue #251 crashes because "fill" overtakes "s" and produces malformed tokens, leading to an infinite loop. Avoid this possibility by writing the output tokens into a scratch buffer, and copy the result back at the end.

Additionally, re-allocate the buffers if fill is "close to" the end of the scratch buffer. Here "close to" is a guess at how much space might be required per iteration.

--------------
Here is a proposed fix of #251. The "magic" `+20` is no longer required, but instead there is a magic "how much space is required at the end of the token buffer for each iteration of the while loop?" At least in the future if someone reports that `fill` goes out of bounds of `tmptokens` it is clear what is going on.

Some `TEMPTY` space is still required at the start of `AC.tokens` before the call of `simp3btoken`: `s` and `t` pointers end up in this space even if initially they are moved beyond it.

I don't measure any performance penalty due to the allocations (they are not all that large) but in principle one could introduce a permanent `AC.tmptokens` to serve as scratch space without the allocations.

Any comments?

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2220,6 +2220,63 @@ assert stdout =~ exact_pattern(<<'EOF')
    S1(spectator)(5)
 EOF
 *--#] Issue231 :
+*--#[ Issue251_1
+#-
+Symbol x,a0,...,a3;
+Local F = <a0*x^0>+...+<a3*x^3>;
+Bracket x;
+.sort
+
+Local G =
+	+ 1 / F[1]
+	+ x * (-F[x]) / F[1]^2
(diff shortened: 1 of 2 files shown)


jodavies:
After a bit more benchmarking I still don't see any impact from the additional memory allocation. I think this patch is good to go. In cases where the `+20` threshold for reallocation is not sufficient, the old code would crash also, and this fixes the problematic cases that we know about.

PR #744, opened 2025-11-21
Implement qgraf-comptatible keywords for FORM interface

jodavies:
With this change, the diagram filtering is compatible with qgraf's output.

This passes many tests (to come, for the test suite) of single keywords.

Currently there is a difference, to be resolved, when specifying both of a "dual option". For example, specifying "onepi + onepr" in qgraf results in 0 diagrams. In FORM the options cancel each other, and everything is generated. A change to grcc is required to resolve this.

Also to discuss:
- what should `WITHSYMMETRIZE` do? GRCC has both `GRCC_OPT_SymmInitial` and `GRCC_OPT_SymmFinal` flags

### DIFF
--- a/sources/grcc.cc
+++ b/sources/grcc.cc
@@ -5154,7 +5154,7 @@ Bool MGraph::isOptM(void)
     selfloop  = (mconn->nselfloops > 0);
     multiedge = (mconn->nmultiedges > 0);
     tadblock  = (mconn->na1blocks > 0);
-    block     = (mconn->neblocks == 1);
+    block     = (mconn->neblocks <= 1);
     extself   = (mconn->ne1bridges > 0);
 
     if (opt->values[GRCC_OPT_1PI] > 0) {


jodavies:
Updated:
- print an error and terminate, if conflicting options are specified, for eg `` `onepi_'+`onepr_' ``

Still to do:
- should `` `withsymmetrize_' `` enable both `GRCC_OPT_SymmInitial` and `GRCC_OPT_SymmFinal`? Should we have two FORM keywords to control them separately?
- test `floop`
- due to the way the options are passed, one has to be careful not to give, eg, `` `onepi_' + `onepi_' `` -- I am not sure there is an easy way to avoid this issue without changing how the options work completely.
- `` `withonepisets_' ``, `` `withblocks_' `` "work" in that they change the output, but I don't know if it is correct
- `` `withedges_' `` presumably can be easily crosschecked against qgraf

I have validated against qgraf at 0,1,2,3 loops, with all pairs of non-conflicting filtering keywords (except floop), for:
Phi34:
- phi - phi
- phi - phi phi
- phi phi - phi phi

Qcd:
- qua - qua
- glu - glu
- gho - gho
- qua - qua glu
- glu - glu glu
- glu glu - glu glu

PR #761, opened 2026-01-08
Grcc fixes

jodavies:
Here are a few fixes for the (interface to) grcc.

1. Don't print symmetry factors in the output for `TopologiesOnly_` mode.
2. Don't allow numbers for couplings in Vertex definitions. It produces an empty argument `,,` in the `node_` output, and anyway crashes for quartic vertices. The manual will need an update to reflect this, but it will be re-written as a separate PR.
3. Don't allow negative coupling powers, these were silently ignored before by discarding the sign.

### DIFF
--- a/check/features.frm
+++ b/check/features.frm
@@ -3013,3 +3013,43 @@ TableBase "no212.tbl" open, readonly;
 .end
 assert runtime_error?('Trying to open non-existent TableBase in readonly mode: no212.tbl')
 *--#] tablebase_ro_2 :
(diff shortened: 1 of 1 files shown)


jodavies:
The above disables `topologies_` in a minimal way, so that working out exactly what code to remove is decoupled, potentially from FORM5 release, and we can sort it out in the future.

The check goes in `TestSub`, in case `topologies_` is generated by a pattern match or something like that.

PR #762, opened 2026-01-09
Relax ordering constraints for $-variable ModuleOptions

tueda:
Here is a minimal implementation that relaxes the ordering constraints between $-variable-related `ModuleOption` statements (`sum`, `maximum`, `minimum` and `local` options) and other statements.

Previous discussions: #296, #188.


tueda:
The "Illegal information in number of arguments properties" is now correctly ignored (non-fatal):
```
CF f>=x<=x;
.end
```

PR #799, opened 2026-02-14
perf: copy optimization

jodavies:
Here is an optimisation experiment, replacing all NCOPY/WCOPY macros with memmove (we can't be sure that memory regions never overlap in use of the macro for memcpy). The replacement alone is a negligible performance improvement (tentatively 1%?) but it is hard to detect within the usual run-to-run variation.

The followup commits improve some existing copies within the code by using the macros instead, and moving some conditionals outside of the copies. I identified the expensive copies with a profiler running the Forcer benchmark.

On my system (Ryzen 7900X, Ubuntu 24.04, GCC 13.3.0, tform -w12,), the results for the usual benchmarks are as follows:
| Benchmark | Speedup w.r.t. v5.0.0 |
| --------- | ------- |
| chromatic | 1.05 ± 0.01 |
| color | 1.01 ± 0.01 |
| fmft | 1.03 ± 0.01 |
| forcer | 1.07 ± 0.00 |
| forcer-exp | 1.08 ± 0.01 |
| mass-fact | 1.00 ± 0.05 |
| mbox1l | 1.01 ± 0.02 |
| minceex | 1.07 ± 0.02 |
| mincer | 1.00 ± 0.05 |
| sort-disk | 0.98 ± 0.02 |
| sort-large | 0.99 ± 0.01 |
| sort-small | 1.01 ± 0.01 |
| trace | 1.02 ± 0.01 |

### DIFF
--- a/sources/declare.h
+++ b/sources/declare.h
@@ -69,11 +69,18 @@
 
 /* (n) is necessary here, since the macro is sometimes passed dereferenced pointers for n */
 #define NCOPY(s,t,n) { while ( (n)-- > 0 ) { *s++ = *t++; } }
-/*#define NCOPY(s,t,n) { memcpy(s,t,n*sizeof(WORD)); s+=n; t+=n; n = -1; }*/
 #define NCOPYI(s,t,n) { while ( (n)-- > 0 ) { *s++ = *t++; } }
 #define NCOPYB(s,t,n) { while ( (n)-- > 0 ) { *s++ = *t++; } }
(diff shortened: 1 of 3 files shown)


jodavies:
I ran benchmarks with more samples, I think indeed the use of memmove doesn't lead to a measurable performance difference. I cleaned up the commits to include only the obvious wins and use the original macros.

This results in 6-7% improvement for Forcer, 3% for mincer-exact, 0-1% for everything else.

PR #814, opened 2026-03-30
fix: various problems in TakeContent and related functions

jodavies:
TakeContent did not correctly remove function content from the terms, it got stuck in infinite loops in several cases. MergeDotproductLists got stuck in infinite loops, missed tail cases, and produced a mal-formed result.

In some of these cases, the poly class proceeded to compute a gcd anyway (since the non-symbol objects are converted to extrasymbols). The flint interface Terminated, since it anticipates content-free input and checks this.

### DIFF
--- a/sources/ratio.c
+++ b/sources/ratio.c
@@ -1167,7 +1167,6 @@ WORD *GCDfunction3(PHEAD WORD *in1, WORD *in2)
 	AR.SortType = SORTHIGHFIRST;
 	term1 = TermMalloc("GCDfunction3-a");
 	term2 = TermMalloc("GCDfunction3-b");
-
 	confree1 = TakeContent(BHEAD in1,term1);
 	tryterm1 = AN.tryterm; AN.tryterm = 0;
 	confree2 = TakeContent(BHEAD in2,term2);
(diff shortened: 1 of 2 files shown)


jodavies:
This one certainly needs some more eyes on it, and probably more test cases.

sources/ratio.c
```diff
@@ -1526,23 +1530,23 @@ novectors:;
...
 		if ( *t >= FUNCTION ) {
 			if ( functions[*t-FUNCTION].commute ) {
-				if ( tcom == 0 ) { tcom = tstore; }
+				if ( tcom == 0 ) { tcom = tout; }
 				else {
 					for ( i = 0; i < t[1]; i++ ) {
 						if ( t[i] != tcom[i] ) {
 							MLOCK(ErrorMessageLock);
-							MesPrint("GCD or factorization of more than one noncommuting object not allowed");
+							MesPrint("&GCD or factorization of more than one noncommuting object not allowed");
```
tueda: This should be a runtime error rather than a compile-time error, because it occurs during module execution and causes ParFORM to hang.

jodavies:
Rebased and added a few more tests (where the content is not just a single object)

PR #816, opened 2026-03-31
fix: forbid non-dollar names starting with $

### LINKED ISSUE #782: Compiler does not reject declarations of $-prefixed symbols
The following code succeeds, though these symbols can't be used, for example, in defining expressions:
```form
On names;
S $s;
V $v;
I $i;
F $f;
Set $ss;
.end
```
```
FORM 5.0.0-beta.1 (Jan 23 2026, v5.0.0-beta.1-319-gd776835)  Run: Mon Jan 26 17:00:35 2026

 Symbols
   $s
 Indices
   $i
 Vectors
   $v
 Functions
   $f
 Sets
   $ss:
```

jodavies:


tueda:
Strangely, I got the following Valgrind error with
```form
Local $test = 1;
.end
```
```
FORM 5.0.0 (Mar 31 2026, v5.0.0-11-gc982339)     Run: Fri Apr  3 17:19:53 2026
    Local $test = 1;
test.frm Line 1 --> Illegal name for expression
==636271== Invalid read of size 4
==636271==    at 0x404FF3D: DoExpr (comexpr.c:251)
==636271==    by 0x40507EF: CoLocal (comexpr.c:66)
==636271==    by 0x4066E94: CompileStatement (compiler.c:696)
==636271==    by 0x4112B5A: PreProcessor (pre.c:1130)
==636271==    by 0x4154E3D: main (startup.c:1819)
==636271==  Address 0x4 is not stack'd, malloc'd or (recently) free'd
==636271==
```
But the Valgrind check in CI passed...?

jodavies:
Right, I can reproduce that locally. The same thing happens with, for eg,
```
Local test_ = 1;
.end
```
It looks like `AC.ProtoType` is only set properly in the non-error branch -- it would be straightforward to simply return the error immediately upon finding a bad expression name?

jodavies:
Now `DoExpr` returns immediately after finding a bad expression name, resolving the valgrind error.

PR #818, opened 2026-04-14
Properly handle runtime errors in ParFORM

tueda:
This PR implements a runtime error propagation mechanism in ParFORM, allowing us to enable test cases that check for runtime errors, which were disabled because ParFORM would otherwise hang.


tueda:
Added a small cleanup commit to remove remaining MPI2-related code.

PR #824, opened 2026-05-05
[ci] Update GitHub Actions to latest major release

gruenich:
Fix deprecation warnings that Node20.js will stop working in June.

There is one case where `actions/checkout@v1` is used. This will still break in June. I am not aware of an obvious replacement.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -276,7 +276,7 @@ runs:
 
     - name: Install dependencies (Ubuntu)
       if: runner.os == 'Linux' && steps.setup.outputs.container == 'false'
-      uses: awalsh128/cache-apt-pkgs-action@a605dbde2ac49a823c9c87ad58491b51848bf355  # for empty_packages_behavior
+      uses: awalsh128/cache-apt-pkgs-action@v1.6.0  # for empty_packages_behavior
       with:
         # We assume libgmp-dev and zlib1g-dev are always installed.
         packages: >-
@@ -450,7 +450,7 @@ runs:
     - name: Cache FLINT
       id: cache-flint
       if: steps.setup.outputs.build_flint == 'true'
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         path: lib/flint
         # Here, we set a conservative cache key so that the cache is not reused
@@ -509,7 +509,7 @@ runs:
     - name: Cache FORM library
       id: cache-formlib
       if: steps.setup.outputs.formlib == 'true'
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         path: formlib
         key: ${{ steps.setup.outputs.formlib_key }}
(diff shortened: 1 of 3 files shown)


gruenich:
Additional to the issue with `actions/checkout@v1`, not all dependencies are ready for Node 20 to 24 switch. See https://github.com/awalsh128/cache-apt-pkgs-action/pull/198

.github/actions/setup-build/action.yml
```diff
@@ -276,7 +276,7 @@ runs:
 
     - name: Install dependencies (Ubuntu)
       if: runner.os == 'Linux' && steps.setup.outputs.container == 'false'
-      uses: awalsh128/cache-apt-pkgs-action@a605dbde2ac49a823c9c87ad58491b51848bf355  # for empty_packages_behavior
+      uses: awalsh128/cache-apt-pkgs-action@acb598e5ddbc6f68a970c5da0688d2f3a9f04d05  # v 1.6.0 for empty_packages_behavior
```
tueda: This line can use `awalsh128/cache-apt-pkgs-action@v1` once the upstream `v1` tag has been updated correctly.

By the way, thanks for opening this PR!

gruenich: Do you ask me to change it to `@v1` now? For me this would be the right way, I just wanted to stick to the style already used for third-party actions.

tueda: I think `@v1.6.0` is preferable for now. Ideally, `@v1` and `@latest` should point to `@v1.6.0`, but they haven't been updated yet. We used `@a605dbd...` because, at the time, this commit had not yet been included in any release.

gruenich: I want to give the `@v1` a try. How can the action be tested?

tueda: When using `v1` (= `v1.5.3`), the CI log shows `Warning: Unexpected input(s) 'empty_packages_behavior', valid inputs are ['packages', 'version', 'execute_install_scripts', 'refresh', 'debug']`. Currently, this doesn't result in any failures because the list of packages provided is always non-empty.

gruenich: I see, I created an issue upstream: https://github.com/awalsh128/cache-apt-pkgs-action/issues/199

tueda (approved):
If we do not expect any progress in `awalsh128/cache-apt-pkgs-action` soon,  I think this PR can be merged as-is for now, keeping it at v1.6.0.

gruenich:
@jodavies I think this should be merged.

jodavies:
What is the implication of using `actions/checkout@v1` in the 32-bit container job? That doesn't actually trigger the warning in the github web UI.

gruenich:
Honestly, I don't know.

jodavies:
If the 32-bit test proves an issue, we also have the option to remove it completely. 32bit builds are "unsupported", after all.

PR #826, opened 2026-05-10
Require autoconf 2.69 & Automake 1.11 or newer

gruenich:


tueda:
If we assume autoconf >= 2.69, then it may also be reasonable to assume, for example, automake >= 1.11.

gruenich:
Sure, I added an according commit.

sources/Makefile.am
tueda: This comment about the workaround for Automake < 1.10 can be removed.

sources/Makefile.am
```diff
@@ -172,11 +172,7 @@ form_CFLAGS = $(COMPILEFLAGS)
 form_CXXFLAGS = $(COMPILEFLAGS)
 form_LDFLAGS = $(LINKFLAGS) $(STATIC_LDFLAGS)
 form_LDADD = $(TOOL_LIBS)
-if AUTOMAKE_GE_110
 form_LINK = $(CXXLD) $(form_CXXFLAGS) $(CXXFLAGS) $(form_LDFLAGS) $(LDFLAGS) -o $@
```
tueda: The values of `form_LINK`, `vorm_LINK`, `tform_LINK` and `tvorm_LINK` are exactly the same as those generated by Automake >= 1.10, so these definitions are redundant and can be removed.
(But `parform_LINK` and `parvorm_LINK` still need to be defined because of `MPICXX`.)

gruenich: I removed the variables you mentioned.

sources/Makefile.am
gruenich: Removed, thanks for the hint!

tueda:
Would you prefer to keep these 4 commits as they are?

To me, either of the following would also be reasonable:
- squash the 2 Autoconf-related commits into one, leaving 3 commits in total: one for Autoconf, one for Automake, and one for the INSTALL file
- squash all changes into a single commit

gruenich:
I squashed in into two commits. This way it should be easy to use `git bisect` in case issues occur.

tueda (approved):
If we decide to drop support for older versions of Autoconf and Automake, this PR looks good to me.

For reference:

- [GNU autoconf 2.69 released [stable]](https://lists.gnu.org/archive/html/autoconf/2012-04/msg00041.html) - bootstrapped with Automake 1.11.1

| Ubuntu   | autoconf        | automake            |
| ---------|-----------------|---------------------|
| precise  (12.04 LTS) | 2.68-1ubuntu2   | 1:1.11.3-1ubuntu2   |
| trusty (14.04 LTS)  | 2.69-6          | 1:1.14.1-2ubuntu1   |
| xenial (16.04 LTS)  | 2.69-9          | 1:1.15-4ubuntu1     |
| bionic (18.04 LTS)  | 2.69-11         | 1:1.15.1-3ubuntu2   |
| focal (20.04 LTS)   | 2.69-11.1       | 1:1.16.1-4ubuntu6   |
| jammy (22.04 LTS)   | 2.71-2          | 1:1.16.5-1.3        |
| noble (24.04 LTS)   | 2.71-3          | 1:1.16.5-1.3ubuntu1 |
| resolute (26.04 LTS) | 2.72-3.1ubuntu2 | 1:1.18.1-3build1    |

gruenich:
@jodavies Do you mind having a look?

jodavies:
It is not impossible that people compile and use form on fairly old systems, for eg some old university HPC facility running EL6 or similar. However, since such systems are (long) unsupported, perhaps we can be optimistic and merge this, and see if anyone complains...

gruenich:
They can use the release tar ball or create one on their own on a newer system. The release tar ball contains a configure script and has no dependency to Autoconf or Autotools (thus the version does not matter).

PR #827, opened 2026-05-11
Additional CI tests from hyperform, fmft and mzv-dm

jodavies:
These commits add some more tests to the CI, from the hyperform and fmft packages. Particularly due to the valgrind tests, this slows the CI down substantially, mostly because a lot of the slow tests come in the same "group" (forcer, fmft, and the slowest of the hyperform tests).

Some of the hyperform tests are manually excluded when running without flint, since their results rely on the sign convention of gcd_ when flint is enabled.


.github/actions/setup-build/action.yml
```diff
@@ -552,6 +556,26 @@ runs:
...
+          cd HyperFORM && git checkout 44705844f8585454c70b6702024b357d15e7b2c3 && cd ..
+          mv HyperFORM/src/{declare-hyperform.h,hyperform.h,mzvlow.h} formlib/
+          # this file needs to be copied to check/extra when running the tests
+          cat HyperFORM/check/*.frm | sed 's/\*{{{/*--#[/' | sed 's/\*}}}/*--#]/' > formlib/hyperform.frm
+          rm -f formlib/hyperform-*
+          touch formlib/hyperform-1.0.0
+        fi
+        if ${{ steps.setup.outputs.fmft }} && [ ! -f formlib/fmft-1.0.0 ]; then
+          rm -f formlib/fmft*
+          wget https://raw.githubusercontent.com/apik/fmft/refs/heads/master/fmft.hh -P formlib
```
tueda: Perhaps, this URL should be: `https://raw.githubusercontent.com/apik/fmft/refs/tags/1.0/fmft.hh`

jodavies: OK, just in case

.github/actions/setup-build/action.yml
```diff
@@ -552,6 +556,26 @@ runs:
           wget https://github.com/a-maier/series/releases/download/1.0.0/series.h -P formlib
           touch formlib/series-1.0.0
         fi
+        if ${{ steps.setup.outputs.hyperform }} && [ ! -f formlib/hyperform-1.0.0 ]; then
+          rm -f formlib/{declare-hyperform.h,hyperform.h,mzvlow.h}
+          git clone https://github.com/adamkardos/HyperFORM.git
+          cd HyperFORM && git checkout 44705844f8585454c70b6702024b357d15e7b2c3 && cd ..
```
tueda: For reference, if you prefer, with Git 1.8.5 or later, this line can also be written as `git -C HyperFORM checkout 44705844f8585454c70b6702024b357d15e7b2c3`.

tueda (approved):

tueda:
Other missing files are:

```
.diagrams_cache/*.dat
diagrams.frm
diagrams/diagrams.h
polynomial.frm
```

which can be in a separate PR if you prefer.

jodavies:
Ah, oops, I will add them in to the last of these commits

PR #829, opened 2026-05-14
Reduce size of thread buckets at larger maxtermsize

jodavies:
These commits address the second point of the discussion in #795. With the default buffer sizes nothing changes, but then the size of the threadbuckets increases only logarithmically with maxtermsize, until eventually it increases linearly again to ensure at least `BUCKETMINTERMS=2` terms fit. But this linear increase is much slower than the current behaviour, which is `125*MaxTermSize`.

The compressbuffer can also start small, and be doubled in size only when necessary.

Only `forcer` and `mbox1l` tests in form-bench set a larger-than-default maxtermsize. With 24 workers, the forcer allocation (80K maxtermsize) goes from 22.4GB -> 20.4GB, but the mbox1l (3M maxtermsize) reduces from 389GB -> 260GB. There is no measurable change in performance.

Any thoughts on this?

### DIFF
--- a/sources/ftypes.h
+++ b/sources/ftypes.h
@@ -995,6 +995,9 @@ typedef int (*TFUN1)(UBYTE *,int);
 // before potentially yielding the block to a waiting reading thread.
 #define MINWRITENUMBEROFTERMS 1
 
+// The minimum number of terms which should fit in a bucket
+#define BUCKETMINTERMS 1
+
 #endif
 
 /*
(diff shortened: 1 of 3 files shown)


jodavies:
Related: #831 separately takes the mbox1l test allocation to 353GB. The forcer test is unaffected (its configured small+large is not the limiting factor).

jodavies:
This doesn't affect the performance of any of the updated ibp-reduction benchmarks, and `ibp-pent1l` (which sets a 7M maxtermsize) allocates 289GB at -w24 instead of 611GB.

jodavies:
Removed some stray debugging prints, and reduced BUCKETMINTERMS to 1.

PR #830, opened 2026-05-14
test: add coverage-increasing tests

jodavies:
Add some tests which cover previously-untested parts of the code, written by inspecting poorly-covered files according to LCOV.

@cbmarini would like to add to this file, so let's merge it as-is and then create more PRs to add further tests.

### DIFF
--- a/check/coverage.frm
+++ b/check/coverage.frm
@@ -1,0 +1,672 @@
+#ifndef `TEST'
+  #message Use -D TEST=XXX
+  #terminate
+#else
+  #include `NAME_' # `TEST'
+#endif
+.end
+
+*--#[ cov_dollar :
+#-
+Off stats;
+#$Pzero = 2-1-1;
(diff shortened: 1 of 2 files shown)


check/coverage.frm
```diff
@@ -0,0 +1,438 @@
...
+    f(2)
+    f(3)
+    f(4)
+    f(5)
+    f(6)
+    f(7)
+    f(8)
+    f(9)
+    f(10)
+EOF
```
tueda, line 574: GPT 5.5 pointed out that this doesn't check for no extra audit output after the expected output.
The next line `    .end` can be added to the `exact_pattern`.

tueda (commented):
Currently, many test files (including `coverage.frm`) are missing from the source tarball because we haven't updated `EXTRA_DIST` in `check/Makefile.am`.

PR #838, opened 2026-06-04
Improve flint performance with multivariate polyratfun

jodavies:
The first commit improves a few hotspots in the flint interface and symbol comparison and normalization.

The second uses FORM's symbol ordering for polyratfun operations, instead of sorting the variables by degree, which the poly class does. This means we can avoid sorting flint's result, since it is already sorted correctly. This leads to large improvements in multivariate polyratfun benchmarks. Speedup w.r.t master:

| Benchmark | 1st commit | 2nd commit |
|:---|---:|---:|
| ibp-box2l  | 1.04 ± 0.02 | 1.09 ± 0.02 |
| ibp-mbox1l | 1.04 ± 0.04 | 1.17 ± 0.04 |
| ibp-pent1l | 1.07 ± 0.05 | 1.32 ± 0.06 |
| ibp-tri3l  | 1.01 ± 0.02 | 1.04 ± 0.01 |

(These are updated ibp benchmarks in the style of "mbox1l", which are not yet merged into form-bench. They have coefficients which are rational polys in 3, 5, 6, 2 variables respectively.).

The caveat is that the signs of the numerator and denominator may change in some cases w.r.t the old behaviour. I would argue that the new behaviour is more consistent, but nonetheless the second commit also adds `On OldPRFSign;` in case the user really needs output which agrees exactly with the old behaviour.

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -1295,6 +1295,8 @@ assert succeeded?
 assert result("F") =~ expr("1")
 *--#] Issue126 : 
 *--#[ Issue128 :
+* The result check here requires the old prf sign convention:
+On oldprfsign;
 * Rational arithmetic giving pi_
 CF rat;
 PolyRatFun rat;
--- a/doc/manual/statements.tex
+++ b/doc/manual/statements.tex
@@ -3825,6 +3825,9 @@ characters can be restored by turning this variable off. See also page
 statement~\ref{substafactarg}\index{factarg} to the new mode of version 4 or 
 later in which expressions in the argument of the mentioned function are 
 completely factored over the rationals. The default is off.}
+
+\leftvitem{3.5cm}{oldprfsign\index{off!oldprfsign}}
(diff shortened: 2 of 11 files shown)


jodavies:
Here are some new numbers, after further updating the ibp tests in form-bench. Speed-up w.r.t master:

| Benchmark | 1st commit | 2nd commit |
|:---|---:|---:|
| ibp-box2l    | 1.04 ± 0.01 | 1.09 ± 0.01 |
| ibp-mbox1l   | 1.05 ± 0.03 | 1.18 ± 0.04 |
| ibp-npbox2l  | 1.02 ± 0.01 | 1.05 ± 0.01 |
| ibp-nptri3l  | 1.00 ± 0.01 | 1.01 ± 0.02 |
| ibp-pent1l   | 1.05 ± 0.03 | 1.31 ± 0.04 |
| ibp-tri3l    | 1.01 ± 0.01 | 1.01 ± 0.01 |

PR #841, opened 2026-06-08
Update actions/checkout to v6 in 32-bit containers

tueda:
This PR fixes problems that occur when using `actions/checkout@v6` in 32-bit containers.

~FLINT can also be tested if the paths are adjusted.~

For background on the Node.js runtime dependency issue in 32-bit containers, see:
- https://github.com/actions/checkout/issues/334
- https://github.com/EliahKagan/gitoxide/pull/5

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -270,13 +270,20 @@ runs:
           echo "CONFIG_ARGS=$args" >>"$GITHUB_ENV"
         fi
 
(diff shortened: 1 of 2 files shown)


jodavies:
For now, the flint interface assumes that WORD is 32 bits. I don't think we need to go out of our way to get this working on 32bit systems.

tueda:
Then, this PR would be fine in its current form.

PR #843, opened 2026-06-09
fix: give each SORTING struct its own OldPosOut, OldPosIn

jodavies:
During a ground-level stage sort, if a sub-buffer par=1 MergePatches creates a sort file, it resets OldPosOut with `PUTZERO(AN.OldPosOut);`. The value of OldPosOut needed by the ground-level sort is lost, leading to terms being lost due to not finishing the Merge by jumping to ConMer, at the end of MergePatches.

Fix by moving OldPosOut, OldPosIn to the SORTING struct. Now every sort has its own private values.

This is a nasty one: users who have run scripts which they know caused both ground-level stage sorts and very nasty sub-buffer sorts (which created sort files) should be suspicious of the results!

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -4656,6 +4656,131 @@ Symbol x;
 .end
 assert succeeded?
 *--#] PullReq691 :
+*--#[ PullReq843_1 :
+#-
+
+#: TermsInSmall 16
+#: LargePatches 10
+#: FilePatches 16
+
+#: SubTermsInSmall 16
+#: SubLargePatches 10
+#: SubFilePatches 10
+
+Off statistics;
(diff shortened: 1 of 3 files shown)


jodavies:
This also looks good if you test a "nested version" like
```
* Generate enough terms in a sub-buffer sort to create a sort file:
Argument f;
   Identify f(x?) = sum_(j,1,`BLOWUP',f(x)/`BLOWUP');
   Argument f;
      Identify x^i?pos_ = sum_(j,1,`BLOWUP',x^i*i/`BLOWUP');
   EndArgument;
   Identify f(x?) = x;
EndArgument;
```
but it takes too long for the CI.

PR #846, opened 2026-06-14
[ci] Update GitHub action awalsh128/cache-apt-pkgs-action to v1.6.1

gruenich:
Fix another deprecation warning for Node.js 20.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -283,7 +283,7 @@ runs:
 
     - name: Install dependencies (Ubuntu)
       if: runner.os == 'Linux' && steps.setup.outputs.container == 'false'
-      uses: awalsh128/cache-apt-pkgs-action@v1.6.0  # for empty_packages_behavior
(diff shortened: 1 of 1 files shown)


.github/actions/setup-build/action.yml
```diff
@@ -283,7 +283,7 @@ runs:
 
     - name: Install dependencies (Ubuntu)
       if: runner.os == 'Linux' && steps.setup.outputs.container == 'false'
-      uses: awalsh128/cache-apt-pkgs-action@v1.6.0  # for empty_packages_behavior
+      uses: awalsh128/cache-apt-pkgs-action@v1.6.1  # for empty_packages_behavior
```
tueda: The comment about `empty_packages_behavior` was originally added because the action was pinned to a commit SHA that did not correspond to any release tag. Now that we are using the latest release tag, the comment is no longer necessary and can be removed.

gruenich: You are right, thanks for the hint. I updated the commit accordingly.

tueda (approved):

PR #848, opened 2026-06-17
test: add #require flint? for the test suite.

cbmarini:
This adds the instruction `#require flint?` such that tests that depend on code that uses the Flint library are skipped if Form is build without it. At the moment it is tested in #845, so if we merge it from this PR I can rebase #845 and remove the commit there.


tueda (approved):
If you would like to add a FLINT-specific property `flint?`, I think the implementation would be fine.

(The [original idea](https://github.com/form-dev/form/issues/645) was `feature?("flint")`, which can be implemented by converting the `-vv` output into a set of available features and checking whether "flint" is included in that set, but this can be added later.)

PR #850, opened 2026-06-24
Fix few resource leaks, uninitialized vars, and error paths

magv:
These small issues are reported by the cppcheck static analyzer; with these changes, cppcheck is happy.


magv:
I've also converted some 0 into NULL in the few places I've touched, if you don't mind.

tueda (approved):
These are indeed improvements. Thanks!

Though it doesn't matter because FORM exits anyway sooner or later in these error paths, it seems that there are more resource leaks around the code you improved. Would you like to address them in this PR as well? Or a separate PR? (Then we'll have another competition in the workshop.)

magv:
No, I think I'm satisfied with this one. Other changes will be separate pull requests.

PR #851, opened 2026-06-24
Hide code from coverage

cbmarini:
This PR wraps unfinished features into LCOV_EXCL_START and LCOV_EXCL_STOP to hide them from coverage statistics. At the moment I hid the (unfinished) code for
- `#namespace`,
- `#setflag`,
- `MultiBracket`,
- the complete `diagrams.c` file.

What else could be hidden because it is not active at the moment? For example the code in `checkpoint.c`?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -216,7 +216,7 @@ jobs:
 
       - name: Generate LCOV coverage data
         run: |
-          lcov -d . -c -o coverage.lcov
+          lcov --config-file .lcovrc -d . -c -o coverage.lcov
           lcov -r coverage.lcov '*/usr/include/*' '*/extern/*'${{ matrix.flint != 'no' && ' ''*/lib/*''' || '' }} -o coverage.lcov
 
       - name: Coveralls Parallel
--- a/.lcovrc
+++ b/.lcovrc
@@ -1,0 +1,8 @@
+# FORM-specific LCOV exclusion markers:
+# 
+# - INTERNAL_ERROR_EXCL_...: exclude internal errors from the coverage statistics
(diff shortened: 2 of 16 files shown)


tueda:
`SkipName` and `ConstructName` are called when reading a model for the diagram generator.

Undocumented user-flag feature is working, except for the `clearflag` statement:
```form
Local F = 1;

#procedure PrintFlag(i,F)
  #if flag(1,F)
    #message flag(1,F) is true
  #else
    #message flag(1,F) is false
  #endif
#endprocedure

#call PrintFlag(1,F)
#setflag 2,F
#call PrintFlag(1,F)
#setflag 1,F
#call PrintFlag(1,F)
#clearflag 1,F
#call PrintFlag(1,F)
setflag 1;
.sort
#call PrintFlag(1,F)
clearflag 1;
.sort
#call PrintFlag(1,F)

.end
```
```
~~~flag(1,F) is false
~~~flag(1,F) is false
~~~flag(1,F) is true
~~~flag(1,F) is false
~~~flag(1,F) is true
~~~flag(1,F) is true
    .end
```
One would expect the last one to be false.

GPT-5.5 suspects that `Add3Com(TYPECLEARUSERFLAG,x)` in `CoClearUserFlag`:
https://github.com/form-dev/form/blob/fffe2088284af458173aaa1f8f0ff4e04f79a0f2/sources/compcomm.c#L7448
should be `Add3Com(TYPECLEARUSERFLAG,x-1)` as in `CoSetUserFlag`.

jodavies:
You could also hide anything to do with the "second stage load balancer", so BalanceRunThread in threads.c and associated code.

Issue #741, opened 2025-11-13
Crash or stuck in FORM-internal loop


nikolai-husung:
The original problem was that I ran into a FORM error message stating:
```
Illegal code in Norm
Illegal code in Norm
Called from TestSub
Program terminating at prepareDimReg Line 103 --> 
Terminate called from proces.c:2156 (TestSub)
Backtrace:
# 0: TerminateImpl at startup.c:1874:10
# 1: TestSub at proces.c:2156:2
# 2: Generator at proces.c:3286:5
# 3: Generator at proces.c:4096:18
# 4: Generator at proces.c:4096:18
# 5: Generator at proces.c:4264:8
# 6: Processor at proces.c:407:11
# 7: DoExecute at execute.c:859:67
# 8: PreProcessor at pre.c:1048:17
# 9: main at startup.c:1750:2
```

While trying to build a minimal example to reproduce the problem I ran into two more variants all related with the replacement in lines 85-88 of the attached FORM script:

(1) I end up in a FORM-internal infinite(?) loop when running the script with -D i=1 or -D i=2 input data.
(2) If I remove in line 88 the overall ft1-factor the script still gets stuck with -D i=1 but crashes with -D i=2 without stating a reason:

```
Program terminating at minError.frm Line 85 --> 
Terminate called from startup.c:1611 (onErrSig)
Backtrace:
# 0: TerminateImpl at startup.c:1874:10
# 1: ?? at startup.c:1611:2
  (inlined by) onErrSig at startup.c:1597:13
# 2: __restore_rt at ??:0
# 3: TestSub at proces.c:705:8
# 4: TestSub at proces.c:1674:13
# 5: Generator at proces.c:3286:5
# 6: Generator at proces.c:4096:18
# 7: Generator at proces.c:4096:18
# 8: Generator at proces.c:4264:8
# 9: Processor at proces.c:407:11
#10: DoExecute at execute.c:859:67
#11: PreProcessor at pre.c:1048:17
#12: main at startup.c:1750:2
```

(3) Specifying further -D workaround runs my current workaround for the problem in lines 75-82 for any of the choices.

For now I only share the "minimal" example. I can try to build a more compact variant of the original code but it is right now rather lengthy. Let me know if you need any more details and thank you for any help with this.

[input1.txt](https://github.com/user-attachments/files/23525281/input1.txt)

[input2.txt](https://github.com/user-attachments/files/23525283/input2.txt)

[...]

PR #874, opened 2026-07-13
fix: function and arg size book-keeping for TestSub recursion

### LINKED ISSUE #367: replace_ fails to handle 0-replacement
The following program causes an error for the replacement of s1 -> 0. For s1-> (other integer) there is no issue. I remember this to work with previous versions.

```
auto s s;
f f I;

L [g] =
      I(f(s1+2,s2,s3))
      ;
multiply replace_(s1,0);

print+s;
.end
```
```
FORM 4.2.1 (Aug 10 2020, v4.2.1-29-g557be9d) 64-bits  Run: Fri Oct 16 14:07:16 2020
    auto s s;
    f f I;

    L [g] =
          I(f(s1+2,s2,s3))
          ;
    multiply replace_(s1,0);

    print+s;
    .end
Internal error in InFunction: Function not encountered.
Program terminating at replacetest.frm Line 9 -->
  0.00 sec out of 0.00 sec
```

jodavies:
When TestSub is called recursively, correctly update the nested sizes. Rely on the recursion to fix the sizes at each level, except when an argument term disappears completely; then we fix up the whole nesting stack and continue processing further argument terms.

This fix was prepared with the help of an LLM report: https://gist.github.com/tueda/9d71a87e04dcc1f51fbbb432ae1af903

Its findings are in line with previous debugging efforts by me, but I did not figure out a proper fix at the time.

In addition:
- completely remove the "subsubveto" mechanism, which appears to be a previous workaround, required due to the incorrect size updates
- add a check that we do not exceed the FunctionLevels setup parameter, and give an error message for the user if so.
- check RecFlag has the same value on entry and when returning from TestSub

---

This fixes #367 and #741

### DIFF
--- a/check/fixes.frm
+++ b/check/fixes.frm
@@ -2863,6 +2863,16 @@ Print;
 assert succeeded?
 assert result("F") =~ expr("f(x**2) + f(x)**2")
 *--#] Issue359 : 
+*--#[ Issue367 :
+* replace_ corruption in nested multi-argument functions
+Function f,I;
+S s1,s2,s3;
+L F = I(f(s1+2,s2,s3))*replace_(s1,0);
+P;
+.end
+assert succeeded?
+assert result("F") =~ expr("I(f(2,s2,s3))")
+*--#] Issue367 :
 *--#[ Issue400 :
 * denominators statement for nested functions
 S x;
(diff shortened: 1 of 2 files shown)


jodavies:
Tentatively, this is also a 1% performance improvement in forcer, forcer-exp and maybe hyperform.

jodavies:
I put back `subsubveto`, since @tueda is not sure about it. I think it would help if we could find a FORM script which would hit the following code, in the case that `AN.subsubveto == 1`:
https://github.com/form-dev/form/blob/b143fc2ba9ba6efbdce8d7c0bda6880a08a636d8/sources/proces.c#L741-L748

jodavies:
While this fixes many examples, I think it the code is still not correct, but for slightly independent reasons. @cbmarini has been also looking at an issue with this same code:
```
CFunction rat;
Symbol ep;
Local F = rat(70944*ep^5,1);
Print "%r";
Multiply replace_(ep,100000);
Print "%r";
Print;
.end
```
Here the issue is the in-place call of `Normalize` here https://github.com/form-dev/form/blob/b143fc2ba9ba6efbdce8d7c0bda6880a08a636d8/sources/proces.c#L1660
In this case, Normalize makes the term larger after raising 100000 to the fifth power: thus Normalize blindly over-writes the trailing term information.

We need to either, copy the term to the workspace and normalize there, and then copy back taking care of the size changes, or not Normalize at all at let the later Normalize during sorting deal with the terms. Then one has to instead make sure that terms which go to 0 are still handled correctly in the Nest stack book-keeping (which was why the Normalize was introduced in the first place).

jodavies:
The second commit resolves the problem by not calling Normalize at all, during the recursion, but relying on the Normalization just before the following sort. The Nest stack is updated entirely by the recursive calls. I am not sure if it is guaranteed that TestSub never increases the size of the term, this does not happen in the test suite at least. The commit adds a Terminate in this case.

As far as I can tell, the code has now come "full circle" back to 2003, except that "subsubveto" is still there?

Any thoughts on this?

jodavies:
This is not yet done: the `Issue741_2` test but with a single replace_, as in
```
Multiply replace_(x,0
   ,y,12345
   ,k1,4321*k2
   ,p1,p2
   ,p3,k1-k2
   ,z,1000000);
```
Hits the new `TestSub: term size increased during recursion.` termination. (And of course, it also crashes without the commits of this PR).

jodavies:
Unless anyone can think of more tests which should be added, we can merge this one now. This part of the code is much better tested than before, and it fixes many cases which previously gave bad results or crashed.

I don't measure any performance regression.

PR #875, opened 2026-07-14
Added "Largest Term" to stats for On SortVerbose

Emaiks:
Added a new line to the statistics message in case of On SortVerbose that shows the maximal term size needed in the past module. For that the maximal term size was checked in:
- AddWithFloat
- AddCoef
- AddPoly
- StoreTerm
- Normalize (at 2 places: before normalization and after)
 
The value will be compared between threads and printed out. No large check have been done.

### DIFF
--- a/sources/float.c
+++ b/sources/float.c
@@ -1754,6 +1754,7 @@ Finished:
 		MUNLOCK(ErrorMessageLock);
 		Terminate(-1);
 	}
+	if ( **ps1 > S->verbMaxTermSize ) S->verbMaxTermSize = **ps1;
 	return(1);
 }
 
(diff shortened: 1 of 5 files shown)


jodavies:
It seems that Normalize can be called, before the associated sorting struct is properly initialised. The failing tests all look like they use `putinside` or `antiputinside` before any function arguments appear in the input. Maybe those functions should make sure a sub-buffer sorting struct is ready before processing anything, rather than removing your termsize checks from Normalize. I'll take a look.

jodavies:
@Emaiks can you wrap the term-size checks in Normalize with `if ( AT.SS == AT.S0 ) {}`? I think it should suffice to only track the ground-level maximal size.

PR #879, opened 2026-07-16
feat: warning if file setup parameters are ignored

jodavies:
If file setup parameters (#: ...) are provided, but there are non-comment lines preceeding them, they are silently ignored. This is consistent with the manual, but easy to do by accident. Add a warning in this case.

---

This is not perfect, in that it won't catch things like
```
#: MaxTermSize 50K
#define TEST "10"
#: SmallSize 1G
```
but it will catch (probably) most instances of this.


jodavies:
Updated version, which also warns in cases like
```
#: MaxTermSize 50K
#define TEST "10"
#: SmallSize 1G
```

PR #884, opened 2026-07-31
Update the stale workflow configuration

tueda:
The stale workflow introduced in https://github.com/form-dev/form/pull/853 has completed its initial pass over the existing issues. We can now increase the `operations-per-run` limit from 15 to the default value of 30 (we could instead set it to a higher value, say, 100).

This PR also updates the cron schedule to avoid receiving stale notifications on weekends.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,7 +1,7 @@
 name: Close stale issues
 on:
   schedule:
-    - cron: '30 0 * * *'  # daily
+    - cron: '30 0 * * 1-5'  # Runs daily on weekdays
   workflow_dispatch:
 
 permissions:
@@ -26,4 +26,3 @@ jobs:
           days-before-pr-stale: -1  # Do not touch PRs.
           days-before-pr-close: -1
           remove-pr-stale-when-updated: false
-          operations-per-run: 15  # Keep stale operations low for now.


tueda:
Somehow the `Issue512_3` test failed for [`tform -w2`](https://github.com/form-dev/form/actions/runs/30620847096/job/91124784849?pr=884):
```
===============================================================================
Issue512_3 (fixes.frm:3127) FAILED
===============================================================================
TFORM 5.0.1 (Jul 22 2026, v5.0.1-25-gdccfb0a) 2 workers  Run: Fri Jul 31 09:48:14 2026
    #-

Time =       0.00 sec    Generated terms =      52500
            test         Terms in output =      52500
                         Bytes used      =    1267016
Please increase SmallExtension setup parameter.
Program terminating in thread 1 at 1.frm Line 15 --> 
Terminate called from sort.c:2040 (AddPoly)
  0.63 sec + 0.25 sec: 0.88 sec out of 0.65 sec
===============================================================================
```
Re-running the job succeeded, but then [this Valgrind job](https://github.com/form-dev/form/actions/runs/30620847096/job/91145543495?pr=884) messed up the cache:
```
Clean installing and caching 3 package(s).
    
    Package list:
    - libmpfr-dev=4.2.1-1build1.1
    - libzstd-dev=1.5.5+dfsg2-2build1.1
    - valgrind=1:3.22.0-0ubuntu3
    Writing main packages manifest to /home/runner/cache-apt-pkgs/manifest_main.log...
    done
    
    Clean installing 3 packages...
    E: Failed to fetch mirror+file:/etc/apt/apt-mirrors.txt/pool/main/g/glibc/libc6-dbg_2.39-0ubuntu8.7_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
    E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
    done
```
```
    /usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/form/form --files-from manifest.txt --use-compress-program zstdmt
    Sent 1305 of 1305 (100.0%), 0.0 MBs/sec
    Cache saved with key: cache-apt-pkgs_afbe0d3042c94273051a4cb20c51d1f8
```
This looks like a bug in the APT package caching action: it fails to detect errors from `apt-fast install` and proceeds to save the cache entry. The root cause of the problem is `404 Not Found`, but it also results in a broken cache entry of only 1.3 KB. 

Hopefully, the 404 error is temporary...

tueda:
It looks like the rollout of the new `ubuntu-24.04` image version `20260810.271.1` is done, though I think #886 is still useful for making the CI more robust.

PR #895, opened 2026-08-23
fix: initialize PreTypes sentinel to fix uninitialized read on stray #endif

### LINKED ISSUE #242: Memory error for extra #endif
Valgrind error for an extra `#endif` (of course this is an error of the user)
```
#ifdef `A'
#message A is defined
#endif
#endif
.end
```
```
FORM 4.2.0 (Oct 26 2017, v4.2.0-22-g954024f) 64-bits  Run: Thu Nov 23 17:42:37 2017
    #ifdef `A'
    #message A is defined
    #endif
    #endif
==20384== Conditional jump or move depends on uninitialised value(s)
==20384==    at 0x4B20CA: DoEndif (pre.c:3204)
==20384==    by 0x4AF3B8: PreProInstruction (pre.c:1156)
==20384==    by 0x4AFF20: PreProcessor (pre.c:931)
==20384==    by 0x4E8EE9: main (startup.c:1605)
==20384== 
test.frm Line 5 ==> #endif without corresponding #if
    .end
```

I think I encountered a case that FORM completely ignored an extra `#endif` in a more complicated code yesterday, but now is not reproducible perhaps due to the undefined behaviour that may depends on memory status of the machine.

Exberg:
## Summary

An extra `#endif` (no matching `#if`) made the preprocessor's termination
handlers (`DoEndif`, and the same peek pattern in `DoElse`, `DoElseif`,
`DoEnddo`, `DoEndprocedure`, ...) read `AP.PreTypes[AP.NumPreTypes]` with an
**empty** stack — i.e. `PreTypes[0]`, a sentinel slot that was never
initialized. `Malloc1` wraps `malloc` without zeroing (tools.c), so the
`!= PRETYPEIF` branch depended on heap garbage:

- Valgrind reported `Conditional jump or move depends on uninitialised value(s)` in `DoEndif` (pre.c), exactly as in this issue's report.
- Depending on the garbage value, the stray `#endif` could also be silently ignored (the non-reproducible case mentioned in the issue).

## Fix

Initialize the sentinel at allocation time in `startup.c`:

```c
AP.PreTypes[0] = PRETYPENONE;
```

This mirrors the existing sentinel initialization of the sibling stacks
(`PreIfStack[0] = EXECUTINGIF` in setfile.c, `PreSwitchModes[0] =
EXECUTINGPRESWITCH` in pre.c). `AddToPreTypes` copies `0..MaxPreTypes` when
it grows the array, so the sentinel is preserved across reallocations. No
behavioral change for correctly nested input: every handler pushes before the
slot can be read, and every matching pop restores `NumPreTypes == 0`.

With the fix, the exact reproducer from the issue now fails deterministically
and identically on every run:

```
test242.frm Line 5 ==> #endif without corresponding #if
Program terminating at test242.frm Line 1 -->
```

## Regression test

Added `Issue242` to `check/fixes.frm`: a stray `#endif` must always be
diagnosed (`assert preprocess_error?("#endif without corresponding #if")`).

## Test plan

Built locally on macOS arm64 (`./configure --disable-float && make -C sources form`; MPFR not available on this machine — see note below).

- [x] Issue reproducer: diagnostic now deterministic across 5 repeated runs
- [x] `./check.rb ./sources/form Issue242` — 1 test, 1 assertion, passes
- [x] Full default suite: **543 tests, 54 failures** — every failure is a float-feature test (`#StartFloat`, `mzv_`, `Evaluate`, `chop`, `torat`, ...), reproduced **identically** (same 54 tests, diffed) on a pristine `master` worktree built with the same `--disable-float` configuration. They are a pre-existing consequence of building without GMP+MPFR on this machine, unrelated to this change. GitHub CI builds with the full feature set.

Note for reviewers: I could not run the Valgrind matrix locally (Linux-only
runner feature); the fix removes the uninitialized read by construction and
the deterministic test covers the observable behavior.

Closes #242


tueda:
There are a few minor style issues with this test case:

- The test is not placed in issue-number order.
- The comments seem a bit verbose (see the other tests).
- The fold is not saved closed (missing trailing whitespace).

More importantly, the test passes with the current version; I would expect a regression test to fail before the fix.

Perhaps the test suite could be extended to detect Valgrind errors directly, regardless of whether other errors occur.

jodavies:
I could imagine something like an `assert valgrind_clean?` would be suitable, that can be used even for error conditions.

tueda:
#913 implements `valgrind_clean?`.

PR #898, opened 2026-08-27
ci: only run the full test suite if FORM source code changes

jodavies:
There is no need to run the full (time consuming) test suite if the source code has not changed (for eg, updating a README etc). The Deploy workflow at least compiles form and runs `make check`, so this would detect problems due to changes to the build configuration or workflow config files.


jodavies:
Updated: the full suite should also run if `test.yml` itself changes.

PR #903, opened 2026-09-08
Add an extra SortBot and make SortBot routine the default

AnaIPereira:
- Add an extra SortBot
- Create a function for the number of SortBots: GetNumberOfSortBots
- Delete MasterMerge and always use SortBotMasterMerge, making the SortBot routine mandatory; 0 and 1 threads also go in this routine, using the logic at the bottom of the tree in the merging process of the SortBots.


jodavies:
Presumably #902 resolves the 32-bit runner issues.

This change affects only a sub-set of the form-bench tests, those in which the master spends a roughly equal amount of time merging terms and writing output. This appears to mean, "using `PolyFun`"; `PolyRatFun`-based tests spend almost all of their time merging terms, and the regular term mode tests spend almost all of their time writing output.

On my usual system, this means:

speedup w.r.t tform-master -w12

| Benchmark | tform-sb -w12 |
|:---|---:|
| chromatic | 1.03 ± 0.01 |
| sort-2-disk | 1.06 ± 0.01 |
| sort-2-large | 1.09 ± 0.01 |
| sort-2-small | 1.08 ± 0.03 |

Since now all tform thread counts >= 1 use the same merging routine, we could remove the `-w2` tests from the CI for a considerable speedup on github actions.

I also have something in the works for compression of scratch files: this change also will mean that scratch compression runs in parallel to term merging for tform.

Issue #215, opened 2017-07-07
Escaping from stage4 sortings for $-variables?


tueda:
Because of really dangerous memory bugs of #211 for $-variables, I would like to avoid stage4 sortings as much as possible. But I am not sure which parameters are really relevant for that. And it is preferable to save the memory usage. Suppose that I don't have any big local expressions, but have $-variables which become potentially very long polynomials. Does the following setting make sense?
```
* Adopt the default values.
#:filepatches             256
#:largepatches            256
#:largesize          50000000
#:smallextension     20000000
#:smallsize          10000000
#:sortiosize           100000
#:termsinsmall         100000

* Try to avoid stage4 sortings for $-variables.
* Setting the same values as non-sub parameters may save us from
* the FORM's confusion of buffer sizes(?)
#:subfilepatches          256
#:sublargepatches         256
#:sublargesize       50000000
#:subsmallextension  20000000
#:subsmallsize       10000000
#:subsortiosize        100000
#:subtermsinsmall      100000

* Slightly smaller maxtermsize for each monomial in polynomials.
#:maxtermsize           10000
```
which consumes 407MB of virtual memory for an empty program. Or better recommendation?

tueda:
Do other parameters affect on the performance for $-variables? For example, `scratchsize`?

vermaseren:
This is tricky stuff when you use tform.
With largesize tform reserves one largesize for the master and one for the combined workers.
That means that for 32 worker, each worker gets largesize/32.
This is not the case with sublargesize. Each worker gets that amount.
It gets even worse: if you have functions inside functions, you need that stuff twice and if, at the
same time you have a dollar there in the inner function, three times. etc.
Dollars, if I am not mistaken, need one buffer per worker and/or master. Each function level
needs one.
The functions do not need such extreme buffers presumably, because in the end they are
limited by MaxTermSize.
This seems to indicate that one would have to decouple the use of the same buffers for dollars
and function arguments. That is a bit messy and means also that potentially you need more space.
The dollars are really not meant to be on disk. It is already nice that it gets through stage 3 at all.
Same for function arguments.
For now the safest seems to be to disallow stage4 for functions and dollars.
Do not forget that when you need stage4 it is only very rare that the output is not very big
(your example was rather artificial I must say) and dollars need to be in allocated memory.

What the effect of your assignments is you just have to try by running and keeping an eye on
the use of the memory in top. Above are the rules I know, but this all depends probably on more
parameters (like maxtermsize).

[...]

PR #911, opened 2026-09-15
fix: make stage4 files sort-private, enable sub-buffer stage4 sorts

### LINKED ISSUE #211: Unexpected code in ReNumber
Hello,

I was writing a small script to see to what extent I could increase the test coverage in `sort.c` and `compress.c`. I have something like the following:
```
#-

#: TermsInSmall 128
#: LargePatches 16
#: FilePatches 4
*#: SubTermsInSmall 64
#: SubLargePatches 8
#: SubFilePatches 2

Off Statistics;

CFunction f,g;
Symbol x,y;

* 128*16=2048 terms cause a sort of the large buffer to disk.
* multiples of 2048*4=8192 terms cause a stage 4 sort
#define NTERMS "40001"

#define ARGNTERMS "2001"

Local test1 = <f(1)>+...+<f(`NTERMS')>;
Local test2 = g(<f(1)>+...+<f(`ARGNTERMS')>);
.sort

* Cancel all terms, but keep distance so that most terms only cancel in the final sort
Identify f(x?) = f(x) - f(`NTERMS'-x+1);
Argument g;
	Identify f(x?) = f(x) - f(`ARGNTERMS'-x+1);
EndArgument;


Print +s;
.end
```
If one uncomments the `#: SubTermsInSmall 64` setting, `tform` crashes with `Unexpected code in ReNumber`. `form` does not crash, but `test2` contains many f(...) terms that are not inside `g`.

Thanks,
Josh

jodavies:
For a long time, Stage4 sorting in the sub-buffers (function arguments, dollar variables, term environments) has been forbidden due to crashes.

Resolve this by moving the Stage4 file handles into the sorting struct; every sort has its own files.

File-to-file based sorting on a per-term basis will be horrible for performance so print a Warning for the user. But FORM will not simply terminate during a potentially long-running program, when the going gets tough.

This resolves #211 and #215

The fix is basically the same deal as the "nearly stage 4" problem I fixed in https://github.com/form-dev/form/commit/f09853d804f2633c15416bca11b7671872043915


jodavies:
While preparing slides for the last workshop, I also noticed that only one of the stage4 files is ever used. The stagesort ping-pong is between one stage4 file, and the main sort file. So also remove the unused file handle.

jodavies:
At the moment, the warning to the user that a sub-buffer stage sort has occurred is printed a single time. Printing it per-term is too much spam, but a single time is probably not prominent enough in the log file. What is a better balance? Once per module? Once per expression per module? Once per expression per thread per module?

